PR #14, opened 2024-03-19
Ft/firebase

zizouet:
Firebase firestore setup and test, don't forget to add the google services.json to the project before testing


SidonieBouthors (approved):
LGTM, passes on my end with google-services.json added :)

PR #50, opened 2024-03-26
Ft/firestore api

zizouet:
- [x] Basic FirebaseAPI class
- [x] Basic Association API
- [x] Basic Association Entity
- [x] Basic Role Entity
- [x] Basic Event Entity
- [x] Basic User Entity


ortoLover (approved):
The code seems to be working overall. The tests pass locally on my machine, and by comparing with what I have done on the bootcamp it makes sense. It seems ready to branch.

PR #59, opened 2024-04-01
Basic receipt API

### LINKED ISSUE #36: Create receipt API

SekoiaTree:
Contains a basic receipt API. Not complete but requesting a review.

Missing:
- [x] Tests
- [x] Various phases of a receipt (need to think about what the structure is in the database)
- [x] Verify that signatures are okay
- [x] Documentation

Closes #36

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/database/ReceiptsAPI.kt
+++ b/app/src/main/java/com/github/se/assocify/model/database/ReceiptsAPI.kt
@@ -1,11 +1,9 @@
 package com.github.se.assocify.model.database
 
-import android.net.Uri
 import com.github.se.assocify.model.entities.MaybeRemotePhoto
 import com.github.se.assocify.model.entities.Receipt
-import com.google.android.gms.tasks.Task
-import com.google.android.gms.tasks.Tasks
 import com.google.firebase.firestore.FirebaseFirestore
+import com.google.firebase.firestore.QuerySnapshot
 import com.google.firebase.storage.FirebaseStorage
 import java.time.LocalDate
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/assocify/model/database/ReceiptsAPI.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.google.firebase.storage.FirebaseStorage
+import java.time.LocalDate
+
+class ReceiptsAPI(userId: String, basePath: String, storage: FirebaseStorage, db: FirebaseFirestore) : FirebaseApi(userId, db) {
+    override val collectionName: String = "$basePath/receipts"
+    private val userCollection = "$collectionName/$userId/list"
+    private val storageReference = storage.getReference(userCollection)
+    private val dbReference = db.collection(userCollection)
+
+    fun uploadReceipt(receipt: Receipt, onPhotoUploadSuccess: (Boolean) -> Unit, onReceiptUploadSuccess: () -> Unit, onFailure: (Boolean, Exception) -> Unit) {
```
SekoiaTree, line 16: This signature is pretty complex, because there's a lot going on.
- onPhotoUploadSuccess has a parameter indicating whether anything was actually uploaded
- onFailure has an argument indicating whether the failure was in the photo upload or the receipt upload

I'm not in love with this signature for sure.

app/src/main/java/com/github/se/assocify/model/entities/Receipt.kt
```diff
@@ -0,0 +1,26 @@
...
+
+data class Receipt(
+    val uid: String,
+    val date: LocalDate,
+    val cents: Int,
+    val phase: Phase,
+    val title: String,
+    val notes: String,
+    val photo: MaybeRemotePhoto,
+)
```
SidonieBouthors, line 19: I would do the following : 
- rename notes to description (for clarity)
- add the payer of the receipt, can be a String (uid of the user)
- add incoming (boolean) for wether this is an expense or a gain
Note that this would also make it compatible with the version I am currently using :) 
I also have amount (double) instead of cents (int) but I agree storing it as cents may be preferable.
Let me know what you think :)

SidonieBouthors (approved):
LGTM :) Thanks !

PR #60, opened 2024-04-06
Association Utils

ortoLover:
An Util class to do various operations on an association and to modify the association values on the database.

The class has as objective to solve some problems that will probably come often when dealing with associations, like accepting a new member, asking to enter in the association as a new member or get some elements of the Association.
Here are some functionalities we may want from it, and the advancement of them:

- [x] Getter for all the associations or some filtered associations
- [x] Getter for users with specific roles in the association
- [x] Create a new association with a new user associated to it
- [x] Request to enter in an existing association
- [x] Accept a new user that wants to enter in a specific association


app/src/main/java/com/github/se/assocify/model/associations/AssociationUtils.kt
```diff
@@ -0,0 +1,144 @@
...
+
+import androidx.lifecycle.ViewModel
+import com.github.se.assocify.model.database.AssociationAPI
+import com.github.se.assocify.model.entities.Association
+import com.github.se.assocify.model.entities.Event
+import com.github.se.assocify.model.entities.Role
+import com.github.se.assocify.model.entities.User
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class AssociationUtils(
```
sarahbadr17, line 11: Since it's a view model, wouldn't it be better to name this class AssociationModel or something else containing the word model in it?

app/src/androidTest/java/com/github/se/assocify/AssociationUtilsTest.kt
```diff
@@ -0,0 +1,177 @@
...
+import com.google.firebase.firestore.CollectionReference
+import com.google.firebase.firestore.DocumentReference
+import com.google.firebase.firestore.DocumentSnapshot
+import com.google.firebase.firestore.FirebaseFirestore
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mockito
+import org.mockito.Mockito.mock
+
+class AssociationUtilsTest {
```
sarahbadr17, line 19: (if you choose to change the name of your class don't forget to change the name of the test class to be consistent)

app/src/androidTest/java/com/github/se/assocify/AssociationUtilsTest.kt
```diff
@@ -0,0 +1,177 @@
...
+  private lateinit var db: FirebaseFirestore
+  private lateinit var assoApi: AssociationAPI
+  private val documentSnapshot = Mockito.mock(DocumentSnapshot::class.java)
```
sarahbadr17, line 22: It seems that the usage of Mockito.mock might be unnecessary here. You could change the mock function calls by removing the Mockito prefix and simply using mock(...) instead.

app/src/androidTest/java/com/github/se/assocify/AssociationUtilsTest.kt
```diff
@@ -0,0 +1,177 @@
...
+    assert(assoUtilsNewUser.getAssociationName() == "")
+  }
+
+  @Test
+  fun checkAcceptNewUser() {
+    Mockito.`when`(documentSnapshot.exists()).thenReturn(true)
+    Mockito.`when`(documentSnapshot.toObject(Association::class.java)).thenReturn(oldAssoUpdated)
+    Mockito.`when`(documentReference.get()).thenReturn(Tasks.forResult(documentSnapshot))
+
+    val task = Tasks.forResult(documentSnapshot)
```
sarahbadr17, line 94: This variable is never used you can safely delete it.

sarahbadr17 (commented):
Overall it looks good to me. However, I would suggest to add some comments especially inside some tests functions so that your code is easier to read and is more accessible.

app/src/main/java/com/github/se/assocify/model/associations/AssociationUtils.kt
ortoLover: It is not a viewModel, I got confused. I used the functions I wrote for the wrong viewModel to create a class containing utility methods to create/modify/send requests to a certain association

app/src/androidTest/java/com/github/se/assocify/AssociationUtilsTest.kt
ortoLover: You're right gonna fix it asap

app/src/androidTest/java/com/github/se/assocify/AssociationUtilsTest.kt
ortoLover: Thank you!!

sarahbadr17 (approved):
LGTM! You can merge !

PR #62, opened 2024-04-07
UI Select Association

sarahbadr17:


app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
SidonieBouthors, line 1: To go with the rest of our current naming scheme we should probably rename this to SelectAssociationScreen :) We could also move it into a folder with the login screen but I think we can deal with that later.

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
```diff
@@ -0,0 +1,143 @@
...
+import androidx.compose.ui.unit.dp
+
+/**
+ * Screen to select an association
+ *
+ * @param registeredAssociation list of organizations that the user is registered to
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun SelectAssociation(registeredAssociation: List<String>) {
```
SidonieBouthors, line 38: Since the list of associations displayed will later come from the viewmodel, I think we can remove it from the arguments now and have a static list in the meantime. Since I believe we're doing the viewmodel next week this shouldn't be an issue though so up to you.

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
```diff
@@ -0,0 +1,143 @@
...
+                  active = isSearching,
+                  placeholder = { Text(text = "Search an organization") },
+                  trailingIcon = {
+                    Icon(
+                        imageVector = Icons.Default.Clear,
+                        contentDescription = null,
+                        modifier = Modifier.clickable(onClick = { /*TODO: clear the search*/}))
+                  },
+                  leadingIcon = {
+                    /* if (isSearching) {
```
SidonieBouthors, line 70: For now isSearching is fixed, but could this still be uncommented now?

app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
```diff
@@ -0,0 +1,114 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+
+/**
+ * This class represents the SelectAssociationScreen
+ *
+ * @param semanticsProvider the semantics provider
+ */
+class SelectAssociationScreen(semanticsProvider: SemanticsNodeInteractionsProvider) :
```
SidonieBouthors, line 25: (If you rename the actual screen to this, make sure you change it to TestScreen or something here so they don't have the same name :))

SidonieBouthors (commented):
Overall this looks pretty good :)

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
sarahbadr17: Not yet, let's wait for the view model ! So that the search can be properly implemented and thus, we can test it and not lose coverage!

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
sarahbadr17: Yes, this argument will indeed be removed after we implemented the viewmodel. However, for the tests I needed to change the list according to the test and it was the easiest way.

app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
sarahbadr17: Done !

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
sarahbadr17: Done !

app/src/main/java/com/github/se/assocify/ui/screens/selectAssoc/SelectAssociation.kt
SidonieBouthors: Sounds good :)

SidonieBouthors (approved):
LGTM I think this is ready to merge :)

PR #75, opened 2024-04-11
Add basic Entity methods and tests for maintanability

zizouet:
I added some basic methods to all entities including :
- [x] Getter
- [x] Equals
- [x] toString
- [x] hashCode
- [x] Private Attributes
- [x] Entity tests
- [x] Add Junit annotations to tests classes to allow test them all at once with tasks.

I also put the main attributes of entities in private to force going by the getters for better maintanability and reduce developer mistakes.

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/entities/User.kt
+++ b/app/src/main/java/com/github/se/assocify/model/entities/User.kt
@@ -4,6 +4,9 @@ data class User(val uid: String, private val name: String, private val role: Rol
   constructor() : this("", "", Role("pending"))
 
   override fun equals(other: Any?): Boolean {
+    if (other?.javaClass != this.javaClass) {
+      return false
+    }
     if (this === other) return true
     other as User
     return uid == other.uid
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/github/se/assocify/model/entities/User.kt
```diff
@@ -1,5 +1,30 @@
 package com.github.se.assocify.model.entities
 
-data class User(val uid: String, val name: String, val role: Role) {
+data class User(val uid: String, private val name: String, private val role: Role) {
   constructor() : this("", "", Role("pending"))
+
+  override fun equals(other: Any?): Boolean {
```
Polymeth, line 6: Not sure why 'other' is of type 'Any?'. In which cases do we compare a role to some other object ?

Polymeth:
In future we would need to formalize having JavaDoc everywhere, but as we have no JavaDoc anywhere I'm not gonna block the PR for this. TODO for later though. Left one question, looks good otherwise it makes the code better :)

app/src/main/java/com/github/se/assocify/model/entities/User.kt
zizouet: equals is an overriden method which only takes other: Any ? as an argument

Polymeth (approved):
Alright good work ;)

PR #76, opened 2024-04-11
Navigation for Login, Select and Choose Association

sarahbadr17:
This PR is associated with issue. It handles the navigation between login, select and choose association. After logging in, if the user is associated to a user in the database, they go to the home tab. If not they must choose or create an organisation. For that, I had to create a LoginViewModel.


sarahbadr17:
Problem wrong pull request: doesn't correspond to the commits in sab/navigationLogin

PR #86, opened 2024-04-12
Viewmodel and updated UI for Receipts

Polymeth:
This PR adds the viewmodel for the "Receipts" page. It also addresses the UI problems related to the addition of the viewmodel.
- Added a viewmodel for the page
- Added a semi functional searchbar
- Changed the list to handle both user's list and the admin overview of all receipts

However, even if the back-end of the search bar is functional, it does not change the display because we decided to change the way the search acted on the UI. The initial idea, easier, is thus not necessary to add. An issue will be opened #95 for this.


app/src/main/java/com/github/se/assocify/ui/screens/treasury/TreasuryScreen.kt
```diff
@@ -35,16 +36,23 @@ import androidx.compose.material3.Tab
...
 import androidx.compose.material3.Text
+import androidx.compose.material3.TextField
 import androidx.compose.material3.TopAppBarDefaults
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.collectAsState
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.rememberCoroutineScope
+import androidx.compose.runtime.setValue
```
zizouet, line 47: I love your imports

app/src/main/java/com/github/se/assocify/ui/screens/treasury/TreasuryScreen.kt
```diff
@@ -212,17 +266,72 @@ fun TreasuryTab(
 fun TreasuryTopBar(
     onAccountClick: () -> Unit,
     onSearchClick: () -> Unit,
+    viewModel: ReceiptViewmodel
```
zizouet, line 269: perfect implementation better like this ! :)

zizouet (approved):

PR #89, opened 2024-04-12
ci: add building APK

SekoiaTree:
"You must set up your repo with a GitHub workflow to build the app's APK such that, at the end of every Sprint, you can create a new release at the push of a button." - M1

This sets that up.


zizouet (approved):
The workflow works perfectly ! For me :)

PR #90, opened 2024-04-12
Add enum for roles

Mai-LinhC:
Adding an enum "RoleTypes" in the Role entities to have a list of all roles currently possible


Polymeth:
Very useful for role checking! Thanks for the work:)

Polymeth (approved):

PR #98, opened 2024-04-14
Receipt Navigation

SidonieBouthors:
Makes sure the following actions work : 

- [x] Creating a receipt by clicking on the "add" button
- [x] Modifying an existing receipt by selecting it in the receipt tab
- [x] Deleting a receipt

Also some minor visual cleanup, and refactors ReceiptAPI

### DIFF
--- a/app/src/androidTest/java/com/github/se/assocify/screens/ReceiptScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/assocify/screens/ReceiptScreenTest.kt
@@ -11,7 +11,7 @@ import androidx.compose.ui.test.performTextClearance
 import androidx.compose.ui.test.performTextInput
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.assocify.model.CurrentUser
-import com.github.se.assocify.model.database.ReceiptsAPI
+import com.github.se.assocify.model.database.ReceiptAPI
 import com.github.se.assocify.model.entities.Phase
 import com.github.se.assocify.model.entities.Receipt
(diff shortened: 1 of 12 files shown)


app/src/androidTest/java/com/github/se/assocify/screens/ReceiptScreenTest.kt
```diff
@@ -11,7 +11,7 @@ import androidx.compose.ui.test.performTextClearance
 import androidx.compose.ui.test.performTextInput
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.assocify.model.CurrentUser
-import com.github.se.assocify.model.database.ReceiptsAPI
```
SekoiaTree, line 14: I don't think it should be named "ReceiptAPI". "ReceiptsAPI" is more correct because it deals with multiple receipts.

SekoiaTree (changes requested):

app/src/androidTest/java/com/github/se/assocify/screens/ReceiptScreenTest.kt
SidonieBouthors: Then we should probably stick to that for UserAPI and AssociationAPI too :) Let's see what the team thinks

SidonieBouthors: Discussed with the team: were gonna keep everything singular for simplicity :)

SekoiaTree (approved):

PR #100, opened 2024-04-15
CreateAsso Tests

SidonieBouthors:
Fixing some tests and minor changes
- [x] Fix tests
- [x] Refactor packages and names to follow the same convention
- [x] Fix minor code issues


app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
```diff
@@ -8,40 +8,22 @@ import com.github.se.assocify.model.database.UserAPI
...
-          User("8", "julie", Role("")),
-          User("9", "bill", Role("")),
-          User("10", "seb", Role("")))
-
   /*
-   * Sets the name of the association
+   * Sets the name of the association : can be any string
    */
   fun setName(name: String) {
-    // need input sanitization TODO
```
zizouet, line 44: We'll may need input sanitization to avoid too long Association Name or changing name to an Assos that already exists for example :)) TODO Later

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
```diff
@@ -184,7 +143,7 @@ class CreateAssoViewmodel() : ViewModel() {
    * Checks if the association can be saved : the current user is a member, all members have a role and the name of the association is not blank
    */
   fun canSaveAsso(): Boolean {
-    return (_uiState.value.members.find { user -> user.uid == currUser } != null) &&
+    return (_uiState.value.members.any { user -> user.uid == CurrentUser.userUid }) &&
         _uiState.value.members.all { it.getRole().getRoleType() != Role.RoleType.PENDING } &&
```
zizouet, line 147: Bad practice, all verifications in the theme of Roles/Permissions should be handled by methods from the role class.
To implement : a method in role. So the better would be to not compare with Role.RoleType.PENDING each time but to have a function in role which makes this test for you

zizouet:
Overall good but lacks coverage and need to implement the role function in another PR

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
zizouet: The method have been merged in main. It is the method isActiveRole

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
Mai-LinhC: task for sprint 5 : #131

zizouet (approved):
Good !

PR #107, opened 2024-04-18
Login problem and clickable select association

sarahbadr17:
This PR fixes the login problem: the current user association id is always set to something and thus there's no null problem.
Plus, when you click on the association on select association you can directly go to home.

### DIFF
--- a/app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
+++ b/app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
@@ -18,8 +18,7 @@ import com.github.se.assocify.model.entities.Role
 import com.github.se.assocify.model.entities.User
 import com.github.se.assocify.navigation.Destination
 import com.github.se.assocify.navigation.NavigationActions
-import com.github.se.assocify.ui.screens.selectAssoc.DisplayOrganization
-import com.github.se.assocify.ui.screens.selectAssoc.SelectAssociation
+import com.github.se.assocify.ui.screens.selectAssociation.SelectAssociation
 import com.kaspersky.components.composesupport.config.withComposeSupport
 import com.kaspersky.kaspresso.kaspresso.Kaspresso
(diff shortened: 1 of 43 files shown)


app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
```diff
@@ -78,6 +83,7 @@ class SelectAssociationTest : TestCase(kaspressoBuilder = Kaspresso.Builder.with
 
   @RelaxedMockK lateinit var mockCurrentUser: CurrentUser
 
+  @RelaxedMockK lateinit var mockFirebase: Firebase
```
SidonieBouthors, line 86: Is it necessary to mockk Firebase given that the API is already mocked? 
Same question for the mockk current user, it doesn't look like you use either of those in the tests.

app/src/main/java/com/github/se/assocify/ui/screens/login/LoginViewModel.kt
```diff
@@ -18,6 +18,7 @@ class LoginViewModel(private val userAPI: UserAPI, private val navActions: Navig
   /** Updates the userId of the UI state */
   fun updateUser() {
     CurrentUser.userUid = getCurrentUser()!!.uid
+    CurrentUser.associationUid = ""
```
SidonieBouthors, line 21: We'll need to fix this later when we are using Supabase and have the right User/Association linking: make sure to add a task for it in the backlog so we don't forget

SidonieBouthors (commented):
Looks good :) Pretty much ready to merge

app/src/androidTest/java/com/github/se/assocify/screens/SelectAssociationScreenTest.kt
sarahbadr17: Yes, I'll remove it. Thanks!

SidonieBouthors (approved):
All good!

PR #140, opened 2024-04-21
Set association uid to non empty string

sarahbadr17:
There was a problem when entering the treasury page because the association uid was set to an empty string. Since we don't have the relation between a user and multiple association, we can't fetch properly the associationuid from the db. I added another issue for later, when the relations will be implemented to fetch the association uid accordingly.


SekoiaTree (approved):
LGTM

PR #142, opened 2024-04-22
Bugfix : change user to have no getters

Mai-LinhC:
Change the entity User back to having public values and no getters


SekoiaTree (commented):
Could you potentially do the same thing for the Association entity? That way we're done with getters for good.

SekoiaTree (approved):
LGTM. Association entity is done in another PR.

PR #143, opened 2024-04-22
Event entity rework for supabase

zizouet:
Rework of Event entity for Supabase and clearer future implementations


zizouet:
Useless finally

PR #148, opened 2024-04-24
Event Main Screen View Model

ortoLover:
For the event screen, we have a main screen containing the different tabs. This main screen has to be able to support diverse functionalities, like swiping from screens, searching a word (that will later be passed to the different tabs) and go back to the user profile. The objective of this PR is to enable all possible functionalities of the main screen of the event tab so that we can later implement thr viewModels of the different tabs

- [x] Go to the user profile
- [x] Do a search
- [x] Change tabs


app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
```diff
@@ -160,60 +155,63 @@ fun EventTab(selected: Boolean, onClick: () -> Unit, text: String, modifier: Mod
...
+              Icon(imageVector = Icons.Filled.Search, contentDescription = "Search")
+            }
+      })
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun EventSearchTopBar(viewModel: EventScreenViewModel) {
+  val state = viewModel.uiState.collectAsState()
+  SearchBar(
```
Mai-LinhC, line 185: You need a way to dismiss the search bar, otherwise it's very annoying for the user : usually, there's a back arrow as a leading element to cancel the search

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
```diff
@@ -160,60 +155,63 @@ fun EventTab(selected: Boolean, onClick: () -> Unit, text: String, modifier: Mod
...
+      modifier = Modifier.padding(8.dp).testTag("searchBar"),
+      query = state.value.searchQuery,
+      onQueryChange = { viewModel.modifySearchQuery(it) },
+      onSearch = { viewModel.modifySearchingState(false) },
+      active = false,
+      onActiveChange = {},
+      trailingIcon = { Icon(imageVector = Icons.Filled.Search, contentDescription = "Search") }) {}
```
Mai-LinhC, line 192: The trailing icon should either work as "input" for searching (as in, "launch" the search), or as it is usual for search bar, have a cross to delete the whole search :)

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
```diff
@@ -160,60 +155,63 @@ fun EventTab(selected: Boolean, onClick: () -> Unit, text: String, modifier: Mod
...
+            }
+      })
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun EventSearchTopBar(viewModel: EventScreenViewModel) {
+  val state = viewModel.uiState.collectAsState()
+  SearchBar(
+      modifier = Modifier.padding(8.dp).testTag("searchBar"),
```
Mai-LinhC, line 186: Not sure what it's due to, but the searchbar is not centered/doesn't take the whole space, it's a bit on the left of the screen

Mai-LinhC (commented):
(I think I have an outdated version so I'm making an incomplete review to refresh)

app/src/main/java/com/github/se/assocify/ui/screens/event/EventGraph.kt
```diff
@@ -2,13 +2,12 @@ package com.github.se.assocify.ui.screens.event
...
 import com.github.se.assocify.navigation.NavigationActions
 
-fun NavGraphBuilder.eventGraph(navigationActions: NavigationActions) {
-  composable(
-      route = Destination.Event.route,
-  ) {
-    EventScreen(navigationActions)
+fun NavGraphBuilder.eventGraph(navigationActions: NavigationActions, eventAPI: EventAPI) {
+  composable(route = Destination.Event.route) {
+    EventScreen(navigationActions, EventScreenViewModel(eventAPI))
```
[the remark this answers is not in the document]
ortoLover: Ok, glad to have been helpful!!

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
ortoLover: Don't know why, fixed it with a fillMaxWidth

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreenViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+    updateEventsFromDb()
+  }
+
+  private fun updateEventsFromDb() {
+    db.getEvents(
+        { events ->
+          updateFilteredEvents(events)
+          _uiState.value = _uiState.value.copy(events = events)
+        },
+        {})
```
Mai-LinhC, line 24: should do something on error (so the user knnows there's been an error)

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
```diff
@@ -124,6 +118,7 @@ fun EventScreen(
                 onClick = {
                   coroutineRoute.launch {
                     pagerState.animateScrollToPage(EventPageIndex.SCHEDULE.index)
+                    viewModel.switchTab(EventPageIndex.SCHEDULE)
                   }
                 })
           }
```
Mai-LinhC, line 129: careful below there's also treasurypage instead of event

Mai-LinhC (changes requested):

Mai-LinhC (approved):

Mai-LinhC:
Still can't swipe on the whole screen for some reason :/
+ there's currently an error with the DB getting events/tasks, so we'll see how it goes : may need a PR later on, but looks fine for now !

PR #155, opened 2024-04-24
feat: translate ReceiptAPI over to supabase

### LINKED ISSUE #154: Translate Receipt API over to supabase

SekoiaTree:
Closes #154


SidonieBouthors (approved):
There were a few functionality issue we discussed IRL but it all seems to work good now :) Still some specificities because UserAPI is still being ported to Supabase but that will all be good next sprint.
Good job :) That was a thougher one !

PR #158, opened 2024-04-25
Dear lord I pushed the wrong main activity

Polymeth:
In my previous PR, I changed main activity and it actually passed the review. Oupsie.


app/src/main/java/com/github/se/assocify/MainActivity.kt
```diff
@@ -17,7 +17,7 @@ class MainActivity : ComponentActivity() {
       AssocifyTheme {
         // A surface container using the 'background' color from the theme
         Surface(modifier = Modifier.fillMaxSize(), color = MaterialTheme.colorScheme.background) {
-          EventMapScreen()
+          AssocifyApp()
```
ortoLover, line 20: The fix seems good to me.

ortoLover (approved):
Everything seems good to me.

PR #162, opened 2024-04-25
Event taskScreen viewModel

ortoLover:
We have just finished working on the whole event screen viewModel, we have now to work on the viewModel of the task tab. We will thus have to set the task management, like the filtering or the clicking of the tasks.
- [x] Manage the add button
- [x] Manage the click on association button
- [x] Work on the filtering
- [x] Fetch the Tasks using the API


app/src/main/java/com/github/se/assocify/model/entities/Task.kt
```diff
@@ -1,27 +1,28 @@
...
-    val name: String = "testName",
+    val title: String = "testName",
     val description: String = "description",
     val isCompleted: Boolean = false,
-    val startTime: String = LocalDate.now().format(DateTimeFormatter.ofPattern("dd/MM/yyyy")),
+    val startTime: LocalDate = LocalDate.now(),
     val peopleNeeded: Int = 0,
     val category: String = "Committee",
-    val location: String = "Here"
+    val location: String = "Here",
```
sarahbadr17, line 26: Do we really want our Location to be a string? maybe coordinates is better?

sarahbadr17 (changes requested):
The event tab is broken: I have this message: error on loading events.

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
```diff
@@ -36,26 +38,30 @@ import androidx.compose.ui.Alignment
...
- * @param currentTab Current tab to display.
+ * @param viewModel The view model for the event screen.
+ * @param taskListViewModel The view model for the task list.
  */
 @OptIn(ExperimentalFoundationApi::class)
 @Composable
-fun EventScreen(navActions: NavigationActions, viewModel: EventScreenViewModel) {
+fun EventScreen(
+    navActions: NavigationActions,
+    viewModel: EventScreenViewModel,
```
Polymeth, line 62: nit: should have a more explicit name, like eventScreenViewModel or something

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreenViewModel.kt
```diff
@@ -51,14 +85,56 @@ class EventScreenViewModel(private var db: EventAPI) : ViewModel() {
...
+
+  /** Stop performing the search and hide the searchbar */
+  fun deactivateSearch() {
+    when (_uiState.value.currentTab) {
+      EventPageIndex.TASKS -> {
+        taskListViewModel.search("")
+        _uiState.value = _uiState.value.copy(searchQuery = "")
+        modifySearchingState(false)
+      }
+      EventPageIndex.MAP -> {
```
Polymeth, line 108: No search in MapScreen, so I don't know how to deal with this

Polymeth (approved):
Few nit picks but overall good job :)

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreen.kt
ortoLover: Will change now!!

app/src/main/java/com/github/se/assocify/ui/screens/event/EventScreenViewModel.kt
ortoLover: Don't we want to filter the tasks on the map depending on their name? We should talk about this later

PR #167, opened 2024-04-25
Fix the app starting screen

ortoLover:
For the second time today, someone forgot to modify the AppScreen that was modified to test the app. This pull request is to revert the startup screen to it's initial state (from profile page to login page).


Mai-LinhC (approved):
LGTM thanks :)

PR #179, opened 2024-04-27
Overhaul user API, role flow

### LINKED ISSUE #119: Move over UserAPI to Supabase

SekoiaTree:
Fixes #119. Currently incomplete.

This PR moves the User API over to Supabase, and additionally overhauls the role system, as well as applying to an association and such.


app/src/main/java/com/github/se/assocify/model/CurrentUser.kt
```diff
@@ -1,6 +1,9 @@
 package com.github.se.assocify.model
 
+import com.github.se.assocify.model.entities.User
+
 object CurrentUser {
   var userUid: String? = null
   var associationUid: String? = null
+  var user: User? = null
```
Mai-LinhC, line 8: we have to keep either uid or user -> discussed in group
probably need to make another PR for that to merge before this one : this one is already too big to do it here
if not, we at least need to put it in scrum board as task for (if possible) this sprint

app/src/main/java/com/github/se/assocify/model/entities/Role.kt
```diff
@@ -22,22 +22,14 @@ data class Role(val name: String) {
...
     return when (name.lowercase()) {
-      "pending" -> RoleType.PENDING
-      "presidence" -> RoleType.PRESIDENCE
+      "presidency" -> RoleType.PRESIDENCY
       "treasury" -> RoleType.TREASURY
       "committee" -> RoleType.COMMITTEE
       "member" -> RoleType.MEMBER
       "staff" -> RoleType.STAFF
-      else -> RoleType.PENDING
+      else -> RoleType.STAFF
```
Mai-LinhC, line 30: we need to keep a pending role : used for when the user is not yet accepted in asso / when you create the asso and the role is not set yet

app/src/main/java/com/github/se/assocify/model/entities/Role.kt
```diff
@@ -22,22 +22,14 @@ data class Role(val name: String) {
...
-   * Whether the role is an active role
-   *
-   * @return whether the role is an active role
-   */
-  fun isAnActiveRole(): Boolean {
-    return getRoleType() != RoleType.PENDING
-  }
 }
+
+data class PermissionRole(val uid: String, val associationId: String, val type: RoleType)
```
Mai-LinhC, line 35: add a comment on what it is??

app/src/main/java/com/github/se/assocify/model/entities/Role.kt
```diff
@@ -1,19 +1,19 @@
...
+  MEMBER,
+  STAFF
+}
+
 data class Role(val name: String) {
   constructor() : this("")
 
   /*
    * enumeration of all roles, sorted by importance
    */
```
Mai-LinhC, line 16: move comment :')

app/src/main/java/com/github/se/assocify/model/entities/User.kt
```diff
@@ -1,17 +1,15 @@
...
-
-  fun toggleRole(role: String): User {
-    return if (hasRole(role)) {
-      User(uid, name, Role("pending"))
-    } else {
-      User(uid, name, Role(role))
-    }
-  }
+@Serializable
+data class User(val uid: String, val name: String, val email: String? = null) {
```
Mai-LinhC, line 6: why add the email??

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationScreen.kt
```diff
@@ -171,19 +171,17 @@ fun CreateAssociationScreen(
...
-                                },
-                                modifier = Modifier.testTag("roleitem-${role.name}"))
-                          }
+                      RoleType.entries.forEach { role ->
+                        ListItem(
+                            headlineContent = { Text(role.name.uppercase()) },
+                            trailingContent = {
+                              RadioButton(
+                                  modifier = Modifier.testTag("role-${role.name}"),
+                                  selected = false, // state.editMember!!.hasRole(role.name),
```
Mai-LinhC, line 180: need to know what role people have -> see other comments

app/src/main/java/com/github/se/assocify/model/entities/User.kt
```diff
@@ -1,17 +1,15 @@
...
+  constructor() : this("", "")
 }
```
Mai-LinhC, line 8: since you removed the toggle role, there still should be somewhere to (dis)able a role for someone in an association : doesn't need to be done here but needs to be put as a task to do in scrum board, otherwise we might forget

app/src/main/java/com/github/se/assocify/model/entities/User.kt
```diff
@@ -1,17 +1,15 @@
...
+
+data class AssociationMember(
+    val uid: String,
+    val associationId: String,
+    val name: String,
```
Mai-LinhC, line 13: I'm not sure i get what uid/name represent, if not the uid/name of the User ? if so, why not just put a user instead ??

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
```diff
@@ -32,7 +32,7 @@ class CreateAssociationViewmodel(
    * Utility function : Sorts the members by role then name
    */
   private fun sortMembers(memberList: List<User>): List<User> {
-    return memberList.sortedWith(compareBy({ it.role.getRoleType().ordinal }, { it.name }))
+    return memberList.sortedWith(compareBy({ it.name }))
```
Mai-LinhC, line 35: you can't just remove stuff :''''))))))
it's supposed to be first sorted by role (because the roles have different importance), then by name
i get that this PR was already too big and the roles were messy but at least leave the previous state of stuff in comment and add a note somewhere (scrum board...) to do this again once we solve the other problems concerning roles etc

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
Mai-LinhC, line 1: note for later PR : need to redo once roles are clean

Mai-LinhC (commented):
notes when testing on emulator : cant change my name in DB 

will finish reviewing later, i'm at 13/26 files.........
please do smaller PRs as we said </3

app/src/main/java/com/github/se/assocify/model/entities/Role.kt
SekoiaTree: I disagree. Roles are only given to members of the association. When a user doesn't have a role yet, it can be `null` temporarily.

app/src/main/java/com/github/se/assocify/model/entities/User.kt
SekoiaTree: This is the actual data stored in the database. An email is useful for contacting a user outside of the app.

app/src/main/java/com/github/se/assocify/model/entities/User.kt
SekoiaTree: An AssociationMember also has a role. I don't know why I added the name, though.

app/src/main/java/com/github/se/assocify/model/entities/User.kt
SekoiaTree: This is dealt with by AssociationMember.

Mai-LinhC (commented):
Looks pretty good, I need to test on device then it'll be done !

Mai-LinhC:
I can't save receipt anymore (with an account that was already logged in) :(

Mai-LinhC (approved):
Looks all good now, thanks :)

PR #181, opened 2024-04-28
Create an API for the tasks

ortoLover:
We already have done the API for multiple values. We now need to concentrate ourselves in the task API, so that we can implement the map later, and the list of tasks. The basic API should contain:

- [x] An updated version of the tasks
- [x] Get all tasks
- [x] Get a task
- [x] Edit a task
- [x] Add a task
- [x] Delete a task


app/src/main/java/com/github/se/assocify/model/entities/Task.kt
```diff
@@ -1,27 +1,28 @@
...
-    val name: String = "testName",
+    val title: String = "testName",
     val description: String = "description",
     val isCompleted: Boolean = false,
-    val startTime: String = LocalDate.now().format(DateTimeFormatter.ofPattern("dd/MM/yyyy")),
+    val startTime: LocalDate = LocalDate.now(),
     val peopleNeeded: Int = 0,
     val category: String = "Committee",
-    val location: String = "Here"
+    val location: String = "Here",
```
sarahbadr17, line 26: I don't think location should be a string. It should be coordinates no?

sarahbadr17 (commented):
LGTM!

app/src/main/java/com/github/se/assocify/model/entities/Task.kt
ortoLover: We still haven't decided how to store the location. We have thought of doing a class location where we store location, but it's still in discussion. I thought about putting a string temporarly, but we can modify it if you want

sarahbadr17: It's fine for now just add an issue in the backlog to modify this !

sarahbadr17 (approved):
LGTM! good job

PR #184, opened 2024-04-29
Viewmodel for profile page : navigation to association settings

Mai-LinhC:
- [x] Create dummy screens
- [x] Go to members
- [x] Go to roles
- [x] Tests


Polymeth (approved):
Nothing to say, good work boss

PR #185, opened 2024-04-29
Fix the EventAPI

ortoLover:
By implementing the Task screen I noticed that the event API does not work. The goal of this class is to standardize the Event class and make it work

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/database/EventAPI.kt
+++ b/app/src/main/java/com/github/se/assocify/model/database/EventAPI.kt
@@ -9,6 +9,9 @@ import kotlinx.coroutines.launch
 import kotlinx.datetime.LocalDate
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
+import java.sql.Timestamp
+import java.time.OffsetDateTime
+import java.util.UUID
 
 class EventAPI(private val db: SupabaseClient) : SupabaseApi() {
 
@@ -43,8 +46,8 @@ class EventAPI(private val db: SupabaseClient) : SupabaseApi() {
   fun getEvents(onSuccess: (List<Event>) -> Unit, onFailure: (Exception) -> Unit) {
     scope.launch {
       try {
-        val events = postgrest.from(collectionName).select().decodeList<Event>()
-        onSuccess(events)
+        val events = postgrest.from(collectionName).select().decodeList<SupabaseEvent>()
+        onSuccess(events.map { it.toEvent() })
       } catch (e: Exception) {
         onFailure(e)
       }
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/assocify/model/database/EventAPI.kt
```diff
@@ -23,7 +26,20 @@ class EventAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+                .insert(
+                    SupabaseEvent(
+                        uid = event.uid,
+                        name = event.name,
+                        description = event.description,
+                        startDate = event.startDate.toString(),
+                        endDate = event.endDate.toString(),
+                        guestsOrArtists = event.guestsOrArtists,
+                        location = event.location))
+                .decodeAs<SupabaseEvent>()
```
SekoiaTree, line 41: Is there a reason we decode at all? The UID is already known.

app/src/main/java/com/github/se/assocify/model/database/EventAPI.kt
```diff
@@ -101,4 +170,25 @@ class EventAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+      val name: String,
+      val description: String,
+      @SerialName("start_date") val startDate: String,
+      @SerialName("end_date") val endDate: String,
+      @SerialName("guests_or_artists") val guestsOrArtists: String,
+      val location: String
+  ) {
+    fun toEvent() =
+        Event(
+            uid = uid ?: "",
```
SekoiaTree, line 186: `uid` is never null; the default value can be removed

app/src/test/java/com/github/se/assocify/model/database/EventAPITest.kt
```diff
@@ -28,6 +29,8 @@ class EventAPITest {
 
   private var error = false
   private var response = ""
+  private var responseTime = OffsetDateTime.now().toString()
```
SekoiaTree, line 32: I think it would be better to have a consistent response time, such as `"2007-12-03T10:15:30+01:00"` (taken from Oracle docs)

app/src/test/java/com/github/se/assocify/model/database/EventAPITest.kt
```diff
@@ -120,27 +123,26 @@ class EventAPITest {
     error = false
 
     val onSuccess: (String) -> Unit = mockk(relaxed = true)
-    val currentTime = LocalDateTime.now()
+    val currentTime = OffsetDateTime.now()
```
SekoiaTree, line 126: Same as above, depending on the current time seems strange in a test.

SekoiaTree (commented):
Looks good, a few nitpicks/cleanups.

app/src/test/java/com/github/se/assocify/model/database/EventAPITest.kt
ortoLover: Gonna change it asap!

app/src/main/java/com/github/se/assocify/model/database/EventAPI.kt
ortoLover: You're right, I will change it.

SekoiaTree (approved):
LGTM

PR #186, opened 2024-04-29
Budget API structure

zizouet:
- [x]  Budget API Interface
- [x] Budget Entity
- [x] Budget API Structure


app/src/main/java/com/github/se/assocify/model/entities/Budget.kt
sarahbadr17, line 1: For me, you don't need a budget entity since we already have accountingCategories or accountingSubCategories. Look into that first and tell me what you think.

app/src/main/java/com/github/se/assocify/model/database/BudgetAPI.kt
```diff
@@ -0,0 +1,97 @@
...
+    // TODO("Not yet implemented")
+  }
+  /**
+   * Add a budget item
+   *
+   * @param budgetUID the unique identifier of the association
+   * @param budgetItem the budget item to add
+   */
+  fun addBudgetItem(
+      budgetUID: String,
```
sarahbadr17, line 28: Is it necessary?

sarahbadr17 (changes requested):
For me, we you can only work with budgetItems and in the viewmodel of the screen, I can filter each budget Item by its category. Good job !

app/src/main/java/com/github/se/assocify/model/database/BudgetAPI.kt
zizouet: Yes, when you need to add a line to your budget you add a budget item

app/src/main/java/com/github/se/assocify/model/entities/Budget.kt
zizouet: I agree with you i removed it :)

sarahbadr17 (approved):
LGTM! you can merge

PR #187, opened 2024-04-30
Repair Task API

ortoLover:
Even if the tests of Task API passed, the database had problems detecting the Time format used locally. Thus I decided to modify the Task value so that it now uses a LocalDateTime which will make it easier to use it.


SidonieBouthors (approved):
Looks good to me :) We can fix our date utils and such to also use offset dates in receipts later perhaps

Issue #175, opened 2024-04-26
Transfer from GMaps to OSM


Polymeth:
Google Maps is very limitating and we realized EPFL's map was using OSM.

PR #188, opened 2024-04-30
EPFL Map integration

Polymeth:
This PR implements the transition between Google Map and OpenStreetMap for the event's map. More specifically, it integreates the EPFL map from the EPFL Plan's API to the application. 

An important problem is that the swipe gestures for the map also swipes the tab, which is an important issue which doesn't have to be adresses now and which is open here #XX. 

Addresses issue #175.


app/src/main/java/com/github/se/assocify/ui/screens/event/maptab/EventMapScreen.kt
```diff
@@ -1,24 +1,99 @@
...
+  // Sets the tile source ot the EPFL plan tiles
+  val campusTileSource = CampusTileSource(0)
+  val tileProvider = MapTileProviderBasic(context, campusTileSource)
+  val tilesOverlay = TilesOverlay(tileProvider, context)
+  mapView.overlays.add(tilesOverlay)
+
+  return mapView
+}
+
+/** The custom tile source from the EPFL plan API. */
```
ortoLover, line 88: Maybe add a comment for the parameter

ortoLover (changes requested):
Everything seems to be working well except for the scrolling problem. We should find a way to fix it ASAP. Also the tests pass well locally, you should correct them so they pass on the CI. But overall it is a good PR that adds an essential functionality to the project. Well done!!

ortoLover (approved):
Everything is good, good PR, everything seems to work ok.

PR #191, opened 2024-05-01
Implement roles

### LINKED ISSUE #190: Implement roles and invitations

SekoiaTree:
Closes #190.

Implements roles. Still needs to be done:
- [ ] Automatically add the current user to the member list when creating an association
- [ ] Automatically accept the invitation to the just-created association

Edit: the above has been moved to another issue, as this PR would become too large if this was added.

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
+++ b/app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
@@ -8,7 +8,6 @@ import io.github.jan.supabase.SupabaseClient
 import io.github.jan.supabase.postgrest.from
 import io.github.jan.supabase.postgrest.query.Columns
 import io.github.jan.supabase.postgrest.query.Count
-import java.time.LocalDate
 import kotlinx.coroutines.launch
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
@@ -235,7 +234,7 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
     db.from("invited")
         .insert(
             Json.decodeFromString<JsonElement>(
-                """{"user_id": "$userId","role_id": "${role.uid}"}"""))
+                """{"user_id": "$userId","role_id": "${role.uid}", "association_id":  "${role.associationId}"}"""))
   }
 
   /**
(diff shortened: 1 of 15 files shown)


app/src/main/java/com/github/se/assocify/model/database/SupabaseObjects.kt
Mai-LinhC, line 1: Why remove it from association API ? And why only this one and not the other objects ?

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -200,19 +209,119 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+  /**
+   * Invites a user to the association with a specific role.
+   *
+   * @param member the member to invite
+   * @param onSuccess called on success
+   * @param onFailure called on failure
+   */
+  fun inviteUser(member: AssociationMember, onSuccess: () -> Unit, onFailure: (Exception) -> Unit) {
+    inviteUser(member.user.uid, member.role, onSuccess, onFailure)
+  }
```
Mai-LinhC, line 261: Why 2 different inviteUser ? Maybe only have one be public ? not sure why both would be needed from the outside

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -200,19 +209,119 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+    tryAsync(onFailure) {
+      inviteUserSus(userId, role)
+      onSuccess()
+    }
+  }
+
+  /**
+   * Initializes an association with roles and users. It is guaranteed that the roles will be
+   * successfully first, then the users. Will fail if the association doesn't exist in the database,
+   * so it must be created first.
```
Mai-LinhC, line 291: I didn't get what this meant at first. Maybe add the precision like "need to call addAssociation before this" or something??

Mai-LinhC (commented):
Looks pretty good, jsut had a few comments. I'll test on device tomorrow. Overall nice job thank you

app/src/main/java/com/github/se/assocify/model/database/SupabaseObjects.kt
SekoiaTree: Because this object is now used by several APIs, so it can't be private anymore

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
SekoiaTree: Will do!

Mai-LinhC (approved):
Tested it on the app, everything seems to work ! Just add the comment and merge :)

PR #193, opened 2024-05-01
Login Skip

SidonieBouthors:
- [x] Set current assoc when logging in as existing user
- [x] Save user and assoc locally
- [x] Load user and assoc from local save
- [x] Log out

### DIFF
--- a/app/src/androidTest/java/com/github/se/assocify/AppTest.kt
+++ b/app/src/androidTest/java/com/github/se/assocify/AppTest.kt
@@ -66,6 +66,7 @@ class ActivityAppTest : TestCase(kaspressoBuilder = Kaspresso.Builder.withCompos
   fun setup() {
     composeTestRule.setContent { MainActivity() }
   }
+
   @Test
   fun loginStart() {
 
--- a/app/src/androidTest/java/com/github/se/assocify/AppTest.kt
+++ b/app/src/androidTest/java/com/github/se/assocify/AppTest.kt
@@ -9,6 +9,7 @@ import androidx.navigation.compose.rememberNavController
 import androidx.test.ext.junit.runners.AndroidJUnit4
 import com.github.se.assocify.model.database.AssociationAPI
 import com.github.se.assocify.model.database.UserAPI
+import com.github.se.assocify.model.localsave.LoginSave
 import com.github.se.assocify.navigation.Destination
 import com.github.se.assocify.navigation.NavigationActions
 import com.github.se.assocify.ui.screens.login.loginGraph
@@ -20,6 +21,7 @@ import io.github.jan.supabase.createSupabaseClient
 import io.github.jan.supabase.postgrest.Postgrest
 import io.ktor.client.engine.mock.MockEngine
 import io.ktor.client.engine.mock.respondBadRequest
+import io.mockk.mockk
 import org.junit.Rule
 import org.junit.Test
 import org.junit.runner.RunWith
@@ -27,7 +29,8 @@ import org.junit.runner.RunWith
 @Composable
 fun LoginApp() {
   val navController = rememberNavController()
-  val navActions = NavigationActions(navController)
+  val mockLoginSave: LoginSave = mockk(relaxed = true)
+  val navActions = NavigationActions(navController, mockLoginSave)
   val supabaseClient: SupabaseClient =
       createSupabaseClient(BuildConfig.SUPABASE_URL, BuildConfig.SUPABASE_ANON_KEY) {
         install(Postgrest)
@@ -36,7 +39,7 @@ fun LoginApp() {
   val userAPI = UserAPI(supabaseClient)
   val associationAPI = AssociationAPI(supabaseClient)
   NavHost(navController = navController, startDestination = Destination.Login.route) {
-    loginGraph(navigationActions = navActions, userAPI = userAPI, associationAPI = associationAPI)
+    loginGraph(navigationActions = navActions, userAPI = userAPI)
   }
 }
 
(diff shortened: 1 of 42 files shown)


app/src/main/java/com/github/se/assocify/navigation/NavigationActions.kt
```diff
@@ -20,11 +24,23 @@ class NavigationActions(private val navController: NavHostController) {
...
+      }
+    }
+  }
+
+  fun onLogout() {
+    loginSaver.clearSavedUserInfo()
+    navController.navigate(Destination.Login.route) {
+      popUpTo(navController.graph.id) { inclusive = true }
     }
   }
```
ortoLover, line 46: What is the use of the onAuthError function?

app/src/main/java/com/github/se/assocify/ui/screens/chat/ChatScreen.kt
```diff
@@ -21,6 +25,11 @@ fun ChatScreen(navActions: NavigationActions) {
             tabList = MAIN_TABS_LIST,
             selectedTab = Destination.Chat)
       }) {
-        Text(modifier = Modifier.padding(it), text = "Chat Screen")
+        Column(
```
ortoLover, line 28: Is this really useful in this PR. Isn't better to implement this functionality on another PR?

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationGraph.kt
```diff
@@ -10,7 +10,7 @@ import com.github.se.assocify.navigation.NavigationActions
 fun NavGraphBuilder.createAssociationGraph(
     navigationActions: NavigationActions,
     userAPI: UserAPI,
-    associationAPI: AssociationAPI
+    associationAPI: AssociationAPI,
```
ortoLover, line 13: Typo: you added a colon at the end of the line

app/src/main/java/com/github/se/assocify/ui/screens/home/HomeScreen.kt
```diff
@@ -21,6 +25,11 @@ fun HomeScreen(navActions: NavigationActions) {
             tabList = MAIN_TABS_LIST,
             selectedTab = Destination.Home)
       }) {
-        Text(modifier = Modifier.padding(it), text = "Home Screen")
+        Column(
```
ortoLover, line 28: Same as the Chat screen, is it a good idea to clean the various screens in this PR?

app/src/main/java/com/github/se/assocify/ui/screens/profile/ProfileScreen.kt
```diff
@@ -259,7 +260,7 @@ fun ProfileScreen(navActions: NavigationActions, viewmodel: ProfileViewModel) {
 
               // log out button (for everyone)
               TextButton(
-                  onClick = { /*TODO logout*/},
+                  onClick = { viewmodel.logout() },
```
ortoLover, line 263: Since the screen has access directly to the navActions, why don't we call logout directly from the navAction?

app/src/main/java/com/github/se/assocify/ui/screens/selectAssociation/SelectAssociationGraph.kt
```diff
@@ -10,7 +10,7 @@ import com.github.se.assocify.navigation.NavigationActions
 fun NavGraphBuilder.selectAssociationGraph(
     navigationActions: NavigationActions,
     userAPI: UserAPI,
-    associationAPI: AssociationAPI
+    associationAPI: AssociationAPI,
```
ortoLover, line 13: Same as last coma, I think you forgot it here

app/src/main/java/com/github/se/assocify/ui/screens/selectAssociation/SelectAssociationScreen.kt
```diff
@@ -51,7 +51,7 @@ import kotlin.math.min
 fun SelectAssociation(
     navActions: NavigationActions,
     associationAPI: AssociationAPI,
-    userAPI: UserAPI
+    userAPI: UserAPI,
```
ortoLover, line 54: Another comma forgot

app/src/androidTest/java/com/github/se/assocify/AppTest.kt
```diff
@@ -36,7 +43,7 @@ fun LoginApp() {
   val userAPI = UserAPI(supabaseClient)
   val associationAPI = AssociationAPI(supabaseClient)
```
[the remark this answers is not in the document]
SidonieBouthors: True

app/src/main/java/com/github/se/assocify/navigation/NavigationActions.kt
SidonieBouthors: It was an old thing we had for when the authentication fails, but we never actually implemented it. It would be nice to have some kind of error message displayed, but I'll remove it from navActions

app/src/main/java/com/github/se/assocify/ui/screens/chat/ChatScreen.kt
SidonieBouthors: True, I just wanted to clean this up real quick :)

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationGraph.kt
SidonieBouthors: My bad, I had added more arguments but ended up removing them

app/src/main/java/com/github/se/assocify/ui/screens/profile/ProfileScreen.kt
SidonieBouthors: I think in general we should consider not passing navActions to screens and having it only be used from viewmodels, ad this seems better practise, which is why I often do this

ortoLover (approved):
Everything is good now, really happy to have reviewed this great PR!!!

PR #213, opened 2024-05-03
Fix CI

### LINKED ISSUE #212: Fix CI
The CI randomly fails occasionally. Somebody should do something about that.

SekoiaTree:
Closes #212

One of the tests randomly failed in the CI. I believe it was because there was a mockk verification that didn't have a timeout, so if the threads didn't line up, it failed.

I wasn't able to test this locally; the test always passed.


Mai-LinhC (approved):
LGTM, hope it fixes the problem, we'll just have to wait, test and see :/

PR #215, opened 2024-05-03
Modify display time

ortoLover:
I forgot to put the date to a more readable format in the task screen. Now I will try modifying it to have it look better.


SidonieBouthors (commented):
Showing time in hour:minutes is enough, you can remove seconds and milliseconds :) Also don't hesitate to use DateUtils and TimeUtils that have already been defined, so the date and time are in the same format as elsewhere in the app

SidonieBouthors (approved):
Amazing thanks for the fix !

PR #221, opened 2024-05-03
Add a get for the current user's role in the current association

SekoiaTree:
This adds a way to get the current user's role in the current association. It is not very efficient, but we will add caching soon.


zizouet (approved):
LGTM perfect !

PR #236, opened 2024-05-06
BalanceAPI

zizouet:
- [x] Balance API
- [x] Tests


app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
```diff
@@ -10,27 +10,25 @@ import java.time.LocalDate
  * @param amount amount of the item
  * @param tva TVA of the item
  * @param description description of the item
- * @param category category of the item (for example: Logistics Pole)
  * @param date date of payment of the item
```
sarahbadr17, line 13: change docs

sarahbadr17 (commented):
LGTM! Just remember changing the java docs when changing the parameters of a function

sarahbadr17 (approved):
Nice job!

Issue #209, opened 2024-05-03
Fix map position on the tab


Polymeth:
The map is wrongly placed on the tab due to OSMdroid. Has to be fixed

PR #238, opened 2024-05-06
Fix major EPFL map bugs

Polymeth:
This PR addresses #209 and #237 which reported the map passing over the tabs of the event page. Also, the swipe gesture was switching the tab and not moving on the map. These two major issues are fixed in this one PR because they are pretty tied to each other and are "easy" to fix.


zizouet:
There is a problem.
While testing the map i only see the EPFL batiments and nothing else :/

zizouet:
LGTM to me :)

zizouet (approved):

PR #243, opened 2024-05-08
Reduce Complexity

SidonieBouthors:
Reduce the complexity of the code


app/src/main/java/com/github/se/assocify/ui/composables/UserSearchTextField.kt
Mai-LinhC, line 1: need to change the doc according to the new params

Mai-LinhC (changes requested):

Mai-LinhC (approved):
LGTM :)

PR #248, opened 2024-05-09
feat: receipt image loading

### LINKED ISSUE #247: Add loading for receipt images

SekoiaTree:
Closes #247. Hand-tested.


SekoiaTree:
Known issue: when changing the image, then reopening the same receipt, the image doesn't update. This is because the local cache isn't updated with the image, and the remote takes a minute or so to update. This is okay for now, as there is a local cache of one minute.

SidonieBouthors (approved):
Besides the know issue we discussed, everything seems to work fine. We need to make sure to add the known issue and loading for the image (so we can visually see when its loading) to the backlog, as well as adding this to the profile picture. 
Thanks :) Cache isn't easy to deal with

PR #266, opened 2024-05-10
BugFix: Multiple Reload

SidonieBouthors:
When refreshing, some pages were sending a new request to the database each time because the viewmodel was entirely recomposed. To avoid this, I'm creating the viewmodels outside of the screen and passing it as argument, everywhere where this wasn't already the case


SekoiaTree (approved):
LGTM

PR #246, opened 2024-05-08
Budgetviewmodel

sarahbadr17:
Budget viewmodel

PR #268, opened 2024-05-10
Implementation of BudgetViewModel

sarahbadr17:
- [x] navigation between viewmodel and screens
- [x] tests
- [x] refactor other tests and screens (name changed)
- [x] viewmodel for budget: getSubCategory lists
- [x] filter by category

Missing filter by year and VTA
This PR will be pushed in ft/AccountingAPI0, because both branches have dependent code.
I closed PR #246 which indicated that I did some code that I didn't do.


ortoLover (changes requested):
Everything seems to work well. Only problem is that I could not check if the sum of the total works (all the categories of the database are equal to 0) and the dates do not filter, but I don't know if it's a wanted functionality. But overall good PR!!

PR #271, opened 2024-05-10
BalanceItem ViewModel

sarahbadr17:
This PR implements the viewmodel for balance items

- [x] get balance items lists from API according to the subcategory
- [x] handle year filter
- [x] handle status filter
- [x] tests


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/balance/BalanceDetailedViewModel.kt
```diff
@@ -0,0 +1,90 @@
...
+  }
+
+  /**
+   * Handle the year filter
+   *
+   * @param year the year to filter by
+   */
+  fun onYearFilter(year: Int) {
+    _uiState.value = _uiState.value.copy(year = year)
+    Log.d("BalanceDetailedViewModel", "Year filter: $year")
```
ortoLover, line 64: Is the log for debug purposes? In that case, wouldn't it be better to remove it?

ortoLover (approved):
A perfect PR! Spent some time trying to find errors but couldn't find any. I am really happy with it. Happy that this part of the code gets merged so we can finish on some other PR's!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/balance/BalanceDetailedViewModel.kt
sarahbadr17: oupsi

PR #272, opened 2024-05-12
Linking detailed screen to accounting screen

sarahbadr17:
- [x] Filter the list of subCategory by uid in the viewmodel
- [x] Correct name displayed in detailed
- [x] Tests
- [x] Only show amount when there’s a list otherwise message indicating empty list

What is missing for those screens to be fully linked is the amount of the subCategories. For right now, it is 0. It will be set to the amount of the corresponding detailed screen when adding an item to the corresponding category


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -193,19 +180,30 @@ fun AccountingDetailedScreen(
...
+              items(budgetModel.budgetList) {
+                DisplayBudgetItem(budgetDetailedViewModel, it, "displayItem${it.uid}")
+                HorizontalDivider(Modifier.fillMaxWidth())
+              }
+
+              // display total amount
+              if (budgetModel.budgetList.isNotEmpty()) {
+                item { TotalItems(totalAmount) }
+              } else {
+                item { Text("No items for the ${subCategory.name} sheet") }
```
Mai-LinhC, line 203: Maybe the text is not the most appropriate, because it depends on the filters so it's misleading

sarahbadr17: Well noticed !

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetDetailedGraph.kt
Mai-LinhC, line 1: After discussion, we decided to leave as is, but if we ever decide to refactor, we shouldn't need a balanceDetailedVM (see how to handle it then : have them nullable, do a closed union...)

Mai-LinhC (approved):
LGTM 💯

PR #273, opened 2024-05-12
Balance View model

sarahbadr17:
This PR refactors the budgetViewModel into AccountingViewModel, because budget and balance sheets have the same list of accountingSubCategory. So they need to have the same view model.


Mai-LinhC (approved):
LGTM :)

PR #274, opened 2024-05-12
Compute the total depending on the TVA filter

ortoLover:
The Budget and balance details screens are almost finished. We just have to compute a total that makes sense, and this is the objective of this task. We will have to do this operation both for the details and the budget. This will only need the creation of an argument for the filter in the view model, and a check for the value on the screen.

- [ ] Implement the filter for the budget
- [ ] Implement the filter for the balance

### DIFF
--- a/app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
+++ b/app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
@@ -61,6 +61,7 @@ import com.github.se.assocify.model.entities.TVA
 import com.github.se.assocify.navigation.NavigationActions
 import com.github.se.assocify.ui.composables.DropdownFilterChip
 import com.github.se.assocify.ui.screens.treasury.accounting.budget.BudgetDetailedViewModel
+import com.github.se.assocify.ui.screens.treasury.accounting.budget.TVAFilter
 import java.time.LocalDate
 
 /**
@@ -142,7 +143,7 @@ fun AccountingDetailedScreen(
 
   val totalAmount =
       when (page) {
-        AccountingPage.BUDGET -> budgetModel.budgetList.sumOf { it.amount }
+        AccountingPage.BUDGET -> if (budgetModel.currentTVAFilter == TVAFilter.TVA_HT) budgetModel.budgetList.sumOf { it.amount } else budgetModel.budgetList.sumOf { it.amount /*TODO: amount if the TVA is applicable*/}
         AccountingPage.BALANCE -> filteredBalanceList.sumOf { it.amount }
       }
 
@@ -188,7 +189,7 @@ fun AccountingDetailedScreen(
               }
               // TODO: change amount given TVA
               DropdownFilterChip(tvaList.first(), tvaList, "tvaListTag") {
-                // TODO: budgetDetailedViewModel.onTVAFilter(it)
+                budgetDetailedViewModel.modifyTVAFilter(TVAFilter.valueOf(it))
               }
             }
           }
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -85,12 +85,22 @@ fun AccountingDetailedScreen(
...
 
   val totalAmount =
       when (page) {
-        AccountingPage.BUDGET -> budgetModel.budgetList.sumOf { it.amount }
-        AccountingPage.BALANCE -> balanceModel.balanceList.sumOf { it.amount }
+        AccountingPage.BUDGET ->
+            if (!budgetModel.filterActive) budgetModel.budgetList.sumOf { it.amount }
+            else
+                budgetModel.budgetList.sumOf {
+                  it.amount + (it.amount * it.tva.rate / 100f).toInt()
```
zizouet, line 96: I think you should display it as a double not a rounded number

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -85,12 +85,22 @@ fun AccountingDetailedScreen(
...
+                }
+        AccountingPage.BALANCE ->
+            if (!balanceModel.filterActive) balanceModel.balanceList.sumOf { it.amount }
+            else
+                balanceModel.balanceList.sumOf {
+                  it.amount + (it.amount * it.tva.rate / 100f).toInt()
```
zizouet, line 102: Same here :)

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingScreen.kt
```diff
@@ -56,7 +56,14 @@ fun AccountingScreen(
         DisplayLine(it, "displayLine${it.name}", page, navigationActions)
         HorizontalDivider(Modifier.fillMaxWidth())
       }
-      item { TotalLine(totalAmount = subCategoryList.sumOf { it.amount }) }
+      item {
+        TotalLine(
+            totalAmount =
+                subCategoryList.sumOf {
+                  if (!model.filterActive) it.amount
+                  else it.amount + it.amount /*TODO: have to find the TVA*/
```
zizouet, line 64: If you leave it like this it is preferred to let an issue in the scrum board

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
ortoLover: I cannot do this. We talked about it, and we need an Int since we are working with money and cannot represent them as floats (too risky). The integer represents the money as cents, so it makes sense to keep them as that (the lowest possible value).

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
ortoLover: Same as up

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingScreen.kt
ortoLover: I'll leave it but will also add an issue to the scrumboard. Sarah should take care of this, she told me so.

zizouet (commented):
kjnkjnk

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
ortoLover: Ok gonna change it asap!!

zizouet (approved):
Tested on my phone and LGTM !! :))

PR #277, opened 2024-05-12
Edit Balance UI

ortoLover:
Now that the edit budget page is done, we have to work on the edit balance part. It should be similar, just having different fields and needed the budgetDetailsViewModel.

- [x] Do the UI
- [x] Do the functionality


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/balance/DisplayEditBalance.kt
```diff
@@ -0,0 +1,314 @@
...
+            item {
+              var tvaExpanded by remember { mutableStateOf(false) }
+              ExposedDropdownMenuBox(
+                  expanded = tvaExpanded,
+                  onExpandedChange = { tvaExpanded = !tvaExpanded },
+                  modifier = Modifier.testTag("categoryDropdown").padding(8.dp)) {
+                    OutlinedTextField(
+                        value = tvaString + "%",
+                        onValueChange = {},
+                        label = { Text("Tva") },
```
Polymeth, line 194: nit nit nit: VAT*

Polymeth (approved):
Works great for me :) Well commented code which is very appreciable!!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/balance/DisplayEditBalance.kt
ortoLover: Mai linh is already doing a PR covering this. It should be fixed soon!!

PR #278, opened 2024-05-13
Cache system and small bugfix for the map

Polymeth:
The map was extremely slow to load. This PR adds pre-fetching for the EPFL campus and a better caching and tiles downloading system.
It also fixes another swap priority bug.

This issues addresses #XX


zizouet:
Map is not loaded for my phone.
To reproduce : 
- Take my phone
- go to the map
- map doesn't load

zizouet (approved):
Perfect PR ! LGTM ! <33

PR #283, opened 2024-05-13
Schedule ViewModel

SidonieBouthors:
- [x] Show tasks on schedule
- [x] Filter shown tasks by selected events
- [x] Switch date by arrows functionality
- [x] Switch date by calendar

### DIFF
--- a/app/src/main/java/com/github/se/assocify/ui/screens/event/scheduletab/EventScheduleViewModel.kt
+++ b/app/src/main/java/com/github/se/assocify/ui/screens/event/scheduletab/EventScheduleViewModel.kt
@@ -1,0 +1,32 @@
+package com.github.se.assocify.ui.screens.event.scheduletab
+
+import com.github.se.assocify.model.database.TaskAPI
+import com.github.se.assocify.model.entities.Task
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+class EventScheduleViewModel(
+    private val taskAPI: TaskAPI,
+) {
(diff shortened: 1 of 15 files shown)


Polymeth:
The schedule displayed is not correct for me. It's displayed 2 hours before. Probably a timeslot problem

app/src/main/java/com/github/se/assocify/ui/screens/event/scheduletab/EventScheduleViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+
+import com.github.se.assocify.model.database.TaskAPI
+import com.github.se.assocify.model.entities.Event
+import com.github.se.assocify.model.entities.Task
+import com.github.se.assocify.ui.util.DateUtil
+import java.time.LocalDate
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+class EventScheduleViewModel(
```
Polymeth, line 11: We really need to add documentation now, even to methods with straight forward names

SidonieBouthors: I added documentation :)

SidonieBouthors:
I fixed the offset issue in the schedule :) the same issue was occurring in the task list so I corrected it globally and added a utils class to correctly deal with offset date times and displays

Polymeth:
Perfect! Thanks for the work and the quick fixes :)

Polymeth (approved):

PR #297, opened 2024-05-14
Bugfix : Profile and join asso

Mai-LinhC:
Minor fixes in the navigation from selectAsso when joining and with the list of displayed associations in profile after that


sarahbadr17 (approved):
LGTM! It works for me :)

PR #299, opened 2024-05-14
End to end test 2

Mai-LinhC:


Mai-LinhC:
Blocked because I can't add a receipt : can't mock the picture taking part

SekoiaTree:
(To add to this, there is a bug in MockK that makes it impossible to mock the viewmodel that handles this, making it essentially impossible to do cleanly)

app/src/main/java/com/github/se/assocify/ui/screens/treasury/receiptstab/ReceiptListScreen.kt
```diff
@@ -104,11 +104,20 @@ fun ReceiptListScreen(viewModel: ReceiptListViewModel) {
...
+ */
 @Composable
-private fun ReceiptItem(receipt: Receipt, viewModel: ReceiptListViewModel) {
+private fun ReceiptItem(receipt: Receipt, viewModel: ReceiptListViewModel, mine: Boolean) {
   ListItem(
-      modifier = Modifier.clickable { viewModel.onReceiptClick(receipt) }.fillMaxWidth(),
+      modifier =
+          Modifier.clickable { viewModel.onReceiptClick(receipt) }
+              .fillMaxWidth()
+              .testTag("$mine-${receipt.uid}"),
```
SekoiaTree, line 120: This test tag should be a bit more explicit - "true-29482a-13etc" isn't very clear. A prefix like "receipt-item" would work.

It feels strange to have an argument that's only for testing convenience, but I don't see a better solution

app/src/androidTest/java/com/github/se/assocify/epics/Epic2Test.kt
```diff
@@ -0,0 +1,304 @@
...
+
+      // check that the receipt is correct and change its title
+      onNodeWithTag("titleField").assertIsDisplayed().performClick().performTextInput("-changed")
+      onNodeWithTag("saveButton").performScrollTo().assertIsDisplayed().performClick()
+      onNodeWithTag("ReceiptList").assertIsDisplayed()
+      onNodeWithText("Receipt-1-name-changed").assertIsDisplayed()
+
+      // (shouldn't have access to budget and balance but not implemented yet)
+
+      // add a receipt -- removed because I can't add a picture
```
SekoiaTree, line 263: nit: maybe add a very short explanation of why not?

SekoiaTree (commented):
Looks pretty good! Only two minor comments

app/src/main/java/com/github/se/assocify/ui/screens/treasury/receiptstab/ReceiptListScreen.kt
Mai-LinhC: yeah it's kind of annoying but there's no other way to differenciate the different list items otherwise :/

SekoiaTree (approved):
LGTM!

PR #300, opened 2024-05-14
Edit a subcategory 

sarahbadr17:
In AccountingDetailedScreen, you can now edit the subCategory by clicking on the edit button on the top of the page. It opens a popup where you can change the name, the year and the tag.  You can also delete a subCategory which deletes every balance and budget items associated to it.

- [x] UI
- [x] ViewModel
- [x] Tests
- [x] Refactoring in Budget and Balance API


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -354,3 +396,159 @@ fun DisplayEditBudget(budgetViewModel: BudgetDetailedViewModel) {
...
+      when (page) {
+        AccountingPage.BALANCE -> balanceState.categoryList
+        AccountingPage.BUDGET -> budgetState.categoryList
+      }
+  var name by remember { mutableStateOf(subCategory.name) }
+  var categoryUid by remember { mutableStateOf(subCategory.categoryUID) }
+  var year by remember { mutableStateOf(subCategory.year.toString()) }
+  var expanded by remember { mutableStateOf(false) }
+  var selectedCategory by remember {
+    mutableStateOf(categoryList.filter { it.uid == subCategory.categoryUID }[0].name)
```
SekoiaTree, line 435: I get a crash here if I click on the edit button before everything loads. Additionally, the "No items" text looks wrong before the load. You should display a loading circle until all the data is loaded.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingViewModel.kt
```diff
@@ -51,6 +51,7 @@ class AccountingViewModel(
                     subCategoryList =
                         subCategoryList.filter { it.categoryUID == _uiState.value.selectedCatUid })
           }
+          Log.d("BudgetViewModel", "Subcategories: $subCategoryList")
```
SekoiaTree, line 54: Debug code left in

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingScreen.kt
```diff
@@ -47,7 +48,7 @@ fun AccountingScreen(
 ) {
   val model by accountingViewModel.uiState.collectAsState()
   val subCategoryList = model.subCategoryList
-
+  Log.d("AccountingScreen", "subCategoryList: $subCategoryList")
```
SekoiaTree, line 51: Debug code left in.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -354,3 +396,159 @@ fun DisplayEditBudget(budgetViewModel: BudgetDetailedViewModel) {
...
+                            balanceViewModel.saveSubCategoryBalanceEditing(
+                                name, categoryUid, year.toInt())
+                        AccountingPage.BUDGET ->
+                            budgetViewModel.saveSubCategoryEditingInBudget(
+                                name, categoryUid, year.toInt())
+                      }
+                    },
+                    modifier = Modifier.testTag("editSubCategorySaveButton"),
+                ) {
+                  Text("Save")
```
SekoiaTree, line 549: These buttons should be dialog buttons. From some quick research, it seems like you should use the "confirm button" and the "dismiss button" of AlertDialog.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -354,3 +396,159 @@ fun DisplayEditBudget(budgetViewModel: BudgetDetailedViewModel) {
...
+ * @param balanceViewModel the viewModel of the balance details
+ * @param balanceState the state of the balance
+ * @param budgetState the state of the budget
+ */
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun DisplayEditSubCategory(
+    page: AccountingPage,
+    budgetViewModel: BudgetDetailedViewModel,
+    balanceViewModel: BalanceDetailedViewModel,
```
SekoiaTree, line 414: nit: I don't like having two essentially identical viewmodels. They also have identical functions, but this forces us to have `when`s everywhere. However, you didn't create this system. Perhaps a problem for another PR?

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -354,3 +396,159 @@ fun DisplayEditBudget(budgetViewModel: BudgetDetailedViewModel) {
...
+                  Icon(
+                      Icons.Default.Close,
+                      contentDescription = "Close dialog",
+                      modifier =
+                          Modifier.clickable {
+                                when (page) {
+                                  AccountingPage.BALANCE ->
+                                      balanceViewModel.cancelSubCategoryBalanceEditing()
+                                  AccountingPage.BUDGET ->
+                                      budgetViewModel.cancelSubCategoryEditingInBudget()
```
SekoiaTree, line 469: Nit: the two methods don't have the same naming pattern.

SekoiaTree (changes requested):
A few changes and a crash, but otherwise good!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -152,29 +190,33 @@ fun AccountingDetailedScreen(
...
 
               // display total amount
-              if (balanceModel.balanceList.isNotEmpty()) {
-                item { TotalItems(balanceModel.balanceList.sumOf { it.amount }) }
+              if (balanceState.balanceList.isNotEmpty()) {
+                item { TotalItems(balanceState.balanceList.sumOf { it.amount }) }
               } else {
-                item { Text("No items for the ${subCategory.name} sheet with these filters") }
+                item {
+                  Text("No items for the ${balanceState.subCategory.name} sheet with these filters")
```
SekoiaTree, line 203: This text should be centered, for cleaniness. Same for the other "no items" text below.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
sarahbadr17: Hi, I agree that it's not good practice and not even good code. I looked into implementing closed Unions in Kotlin, which can be done with a sealed class, but the code would have been even more horrendous and it would have cost me too much time for what it is. Maybe we can merge these two viewmodel, since they are practically the same. This can be another Pr, but in my opinion, it's not the priority for our app. We can do this another time for the "v2" of our project. Right now, we need to satisfy the product owners requirements,  and we would spend too much time refactoring our code.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
sarahbadr17: Loading will be done in another PR! But yes, I agree, we shouldn't be able to click on it when the loading has not be done. I can add a condition for that

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
sarahbadr17: As you can see, I'm not using AlertDialog, because it's deprecated. The Dialog composable that I'm using doesn't have any parameters of this kind, so it seems that I don't have any option rather than using these Textbutton.

SekoiaTree: AlertDialog is deprecated, but it has been replaced by BasicAlertDialog, which is *not* deprecated.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
SekoiaTree, line 1: Two small visual notes:
- the amount number is smaller than on the receipt screen. You should add a `style = MaterialTheme.typography.bodyMedium` to the text
- the balance item composable has some padding on the amount that the budget item doesn't have

SekoiaTree (commented):
Almost good! I noticed some visual stuff while testing that should just be a few lines, if you could fix those it'd be ready in my opinion!

SekoiaTree (approved):
LGTM! Good job!

PR #301, opened 2024-05-14
Allow to move subcategories between categories in DB

zizouet:
PATCH : the last PR that wasn't functional.
Now in the DB it is possible to edit the categoryUID of a subcategory.


sarahbadr17 (approved):
LGTM! You can merge !

PR #304, opened 2024-05-15
Task Duration & Schedule Overlaps

SidonieBouthors:
- [x] Add duration to tasks in DB
- [x] Add duration to Task entity and support in API
- [x] Add possibility to input a duration in task edit
- [x] Add duration to schedule
- [x] Algorithm for task overlaps in the schedule (difficult :'))
- [x] Deal with tasks spanning multiple days
- [x] Add clicking on a task in schedule to edit

Features:
The user can now enter the duration of a Task
The user can now view simultaneous tasks in the schedule without breaking the view
The user can now create tasks that span multiple days and these are correctly displayed in the schedule
The user can now click on a task in the schedule to modify it


app/src/main/java/com/github/se/assocify/ui/screens/event/scheduletab/EventScheduleViewModel.kt
```diff
@@ -91,6 +96,168 @@ class EventScheduleViewModel(
...
+              val currentDayDuration = DateTimeUtil.toLocalTime(endTime).toSecondOfDay() / 60
+              it.copy(
+                  // Set the start time to the beginning of the day
+                  startTime =
+                      DateTimeUtil.toOffsetDateTime(_uiState.value.currentDate, LocalTime.MIN),
+                  // Set the duration to the remaining time of the task
+                  duration = currentDayDuration)
+            }
+
+    val dayTasks =
```
ortoLover, line 130: What about the tasks that are between two days?

ortoLover (commented):
Good PR!! Really complex to review, can't imagine how it was to code. However good code, really impressive algorithms you have wrote! A small nitpick if you have time, you could write some comments on the two algorithms of the view model so that it is easier to understand for posterity. However good job!!

app/src/main/java/com/github/se/assocify/model/entities/Task.kt
```diff
@@ -21,6 +22,7 @@ data class Task(
     val description: String,
     val isCompleted: Boolean,
     val startTime: OffsetDateTime,
+    val duration: Int,
```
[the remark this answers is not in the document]
SidonieBouthors: Both options are fine. When I brought it up most team members seemed to say that just having a start time + duration combo is easier than start time + end time combo. Both have advantages and drawbacks, but we did opt to have the user fill in the duration (more intuitive) so we went for this.

app/src/main/java/com/github/se/assocify/ui/composables/TimePickerWithDialog.kt
```diff
@@ -66,7 +66,10 @@ fun TimePickerWithDialog(
     onTimeSelect: (LocalTime?) -> Unit,
     modifier: Modifier = Modifier,
     label: @Composable (() -> Unit)? = null,
-    errorText: @Composable (() -> Unit)? = null
+    isError: Boolean = false,
+    errorText: @Composable (() -> Unit)? = null,
+    dialogTitle: String = "Select Time",
```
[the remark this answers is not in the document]
SidonieBouthors: Of course :)

app/src/main/java/com/github/se/assocify/ui/screens/event/scheduletab/EventScheduleViewModel.kt
SidonieBouthors: I check tasks that start today (and clamp them to end at midnight if necessary), and also check tasks that end today. I need to double check whether I check tasks that are ongoing today but do not start or end today, thanks!

SidonieBouthors: Done! I added support for tasks that span the whole day without starting or ending on said day :)

ortoLover (approved):
Everything works now good job!!

Issue #206, opened 2024-05-03
Add caching to APIs


SekoiaTree:
- [x] AssociationAPI (handled by #241)
- [x] BudgetAPI (handled by #303)
- [x] BalanceAPI (handled by #305)
- [x] AccountingCategoryAPI (handled by #307)
- [x] AccountSubCategoryAPI (handled by #307)
- [x] EventAPI (handled by #309)
- [x] ReceiptAPI (handled by #302)
- [x] TaskAPI (handled by #310)
- [x] UserAPI (handled by #308)

PR #305, opened 2024-05-15
Balance API caching

SekoiaTree:
Completes part of #206.

This PR adds caching to the balance API, which allows it to load faster. The cache doesn't receive updates from the remote, so it needs to be manually refreshed (currently, by reopening the app). In the future, we will be able to drag down on the tab to refresh the cache.


zizouet (approved):
Perfect LGTM !

PR #307, opened 2024-05-15
Category and subcategory caching

SekoiaTree:

This PR adds caching to the AccountCategoryAPI and AccountSubcategoryAPI. Similarly to other caching PRs, it allows manual refreshing, but not automatic.

### DIFF
--- a/app/src/test/java/com/github/se/assocify/model/database/AccountingCategoryAPITest.kt
+++ b/app/src/test/java/com/github/se/assocify/model/database/AccountingCategoryAPITest.kt
@@ -9,6 +9,7 @@ import io.ktor.client.engine.mock.MockEngine
 import io.ktor.client.engine.mock.respond
 import io.ktor.client.engine.mock.respondBadRequest
 import io.ktor.http.Headers
+import io.mockk.clearMocks
 import io.mockk.junit4.MockKRule
 import io.mockk.mockk
 import io.mockk.verify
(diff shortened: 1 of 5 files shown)


Mai-LinhC (commented):
Only things to note here are out of the scope of this PR : 
- When changing association and going back to the treasury, the categories were not well displayed : there's actually an error in the way the error in loading is handled in balance/budget which has already been notified and added to the backlog (it should be solved once loading is added in these pages)
- There's no caching for what's to be written once internet comes back : that might not be implemented in the app altogether.
Otherwise, looks good to me, pretty neat way of caching the read of categories

Mai-LinhC (changes requested):
The inits are useless here because the API is created before the user infos are loaded, so the assoUID of CurrentUser is null at this point, needs to be moved in AssocifyApp.

app/src/test/java/com/github/se/assocify/model/database/AccountingCategoryAPITest.kt
```diff
@@ -64,6 +65,13 @@ class AccountingCategoryAPITest {
 
     api.getCategories(CurrentUser.associationUid.toString(), onSuccess, onFailure)
     verify(timeout = 400) { onSuccess(any()) }
+    clearMocks(onSuccess)
+
+    error = true
+
+    // Test cache
+    api.getCategories(CurrentUser.associationUid.toString(), onSuccess, onFailure)
+    verify(timeout = 400) { onSuccess(any()) }
```
Mai-LinhC, line 74: no verify for onFailure here, but present in subcategory tests

Mai-LinhC (approved):

PR #308, opened 2024-05-15
User API caching

SekoiaTree:

This PR adds caching to the User API. This makes the profile page and association creation page load faster. The cache needs to be manually refreshed to receive outside updates, but it does not need to be updated manually after local changes.

Tests have not been added, but depending on current coverage I will add some.

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
+++ b/app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
@@ -106,8 +106,9 @@ class UserAPI(private val db: SupabaseClient) : SupabaseApi() {
     tryAsync(onFailure) {
       db.from("users").update({ User::name setTo newName }) { filter { User::uid eq userId } }
 
-      if (userCache[userId] != null) {
-        userCache[userId] = userCache[userId]!!.copy(name = newName)
+      val cachedValue = userCache[userId]
+      if (cachedValue != null) {
+        userCache[userId] = cachedValue.copy(name = newName)
       }
       onSuccess()
     }
--- a/app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
+++ b/app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
@@ -22,24 +22,40 @@ import kotlinx.serialization.json.JsonObject
  */
 class UserAPI(private val db: SupabaseClient) : SupabaseApi() {
 
+  private var userCache = mutableMapOf<String, User>()
+
+  init {
+    updateUserCache({}, {})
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
```diff
@@ -50,9 +66,10 @@ class UserAPI(private val db: SupabaseClient) : SupabaseApi() {
    * @return a list of all users
```
zizouet, line 66: don't return anymore

app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
```diff
@@ -22,24 +22,40 @@ import kotlinx.serialization.json.JsonObject
...
+      onSuccess(users)
+    }
+  }
+
   /**
    * Gets a user from the database
    *
    * @param id the id of the user to get
    * @param onSuccess called on success with the user
-   * @return the user with the given id
```
zizouet, line 30: onfailure here

app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
```diff
@@ -199,12 +246,10 @@ class UserAPI(private val db: SupabaseClient) : SupabaseApi() {
       onSuccess: (List<Association>) -> Unit,
       onFailure: (Exception) -> Unit
   ) {
-    tryAsync(onFailure) {
-      val associations =
-          db.from("member_role_association_view")
-              .select { filter { Membership::userId eq CurrentUser.userUid!! } }
-              .decodeList<Membership>()
-      onSuccess(associations.map { it.getAssociation() })
+    if (currentUserAssociationCache != null && currentUserAssociationId == CurrentUser.userUid) {
```
zizouet, line 249: Here is it normal you compaer userUID with association UID ? :)

app/src/main/java/com/github/se/assocify/model/database/UserAPI.kt
```diff
@@ -185,10 +209,33 @@ class UserAPI(private val db: SupabaseClient) : SupabaseApi() {
...
     tryAsync(onFailure) {
       db.from("users").delete { filter { User::uid eq id } }
+
+      userCache.remove(id)
       onSuccess()
     }
   }
 
+  private var currentUserAssociationCache: Map<String, Pair<Association, PermissionRole>>? = null
+  private var currentUserAssociationId: String? = null
```
zizouet, line 219: I think this name is confusing, makes think about AssociationUID more than user

zizouet (approved):
Excellent ☀️👄☀️

PR #309, opened 2024-05-15
Event API caching

SekoiaTree:

This PR adds caching to the Event API. The cache needs to be manually refreshed to receive outside updates, but it does not need to be updated manually after local changes.

It also replaces all the `scope.launch`es with `tryAsync`, for consistency.

Tests have not been added, but depending on current coverage I will add some.


zizouet (approved):
LGTM good job !

PR #310, opened 2024-05-15
Task caching

SekoiaTree:

This PR adds caching to the Task API. This makes the task page load faster. The cache needs to be manually refreshed to receive outside updates, but it does not need to be updated manually after local changes.


app/src/main/java/com/github/se/assocify/model/database/TaskAPI.kt
```diff
@@ -113,6 +134,22 @@ class TaskAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+          taskCache?.map {
+            if (it.uid == uid) {
+              it.copy(
+                  title = title,
+                  description = description,
+                  isCompleted = isCompleted,
+                  startTime = startTime,
+                  peopleNeeded = peopleNeeded,
+                  category = category,
+                  location = location)
```
SidonieBouthors, line 148: I think duration is missing :)

SidonieBouthors (changes requested):
Overall seems to work when I use the app and go offline and reload the tab :)

SidonieBouthors (approved):
Looks good to me, works on my phone :)

PR #315, opened 2024-05-16
Verify association name on creation

### LINKED ISSUE #131: Association Util : Check if name is valid (not too long, not the same as already existing one ?)

SekoiaTree:
Currently, when associations are created, there is no feedback for the name, but it is impossible to create an association with the same name as another. This PR fixes this, by adding a search through the cache. If the cache is not there, we return to old behavior (i.e. it is still impossible to create that association).

Closes #131

@Mai-LinhC will help me fix the UI looking a bit strange (centering acts weirdly with error messages).


app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -74,6 +74,19 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+   * Checks if the association name is valid. This means that the name is not empty and is not
+   * already taken. This function does not send any requests to the database, because it might be
+   * called multiple times in quick succession. As such, call `updateCache` before calling this if
+   * you want to ensure the cache is up to date.
+   *
+   * @param name the name to check
+   * @return true if the name is valid, false otherwise
+   */
+  fun associationNameValid(name: String): Boolean {
+    return name.isNotBlank() && associationCache.values.none { it.name == name }
```
sarahbadr17: You forgot the case where you add a blank space. They will also be indistinguishable

sarahbadr17 (changes requested):
Nice job! Just make sure that. no can write the same name with a blank space

sarahbadr17 (approved):
Nice job :)

PR #316, opened 2024-05-16
Link SubCategories amount to the items

sarahbadr17:
This PR links the amount of the subCategories present in Budget and Balance to the amount of the items related to each of these categories. The only thing that changes in the budget and balance page is the amount of the subCategory, because what you budget is not the same as what you pay.

In the same PR, the amount in accounting screen can be filtered by TVA. 

I also put the displayEditSubCategory in another file to reduce complexity

I removed the yearFilter in accountingDetails because it isn't necessary

When you add an amount, it shouldn't overflow.

I found a bugFix in the dropdownmenus on the accounting screens: when you change from budget and balance, the filters were reset to the default and the list of subCategory weren't reset. 

I changes some UIs stuff tu make it better


Mai-LinhC:
I don't know if it's due to this PR but when I go to administration (balance) with the right tags, nothing appears at first, and i have to change filters then go back to 2024 and then Poulpe loads :/

Mai-LinhC:
Don't know (yet) if it's from this PR, but the amount is an int, so "10" is displayed as "0.10" and putting a point crashes the app (both in edit balance and edit budget items). but the amount put in the item is well displayed in the sub category and category otherwise. A bit hard to test much since there's no way to add an item here. Again, for both subcategory pages (in balance and in budget), the items don't get displayed at first and i need to reload (change filters then back) for it to load. 
Edit budget item (dialog) is not scrollable which might get problematic on smaller phones. in budget, editing items doesn't actually change them ??? it changes the item i see once i save, but if i leave the page the amount the not there and if i go back to see the items they did not get modified.

sarahbadr17:
> Don't know (yet) if it's from this PR

Not from these PR and now fixed on main!

sarahbadr17:
> I don't know if it's due to this PR but when I go to administration (balance) with the right tags, nothing appears at first, and i have to change filters then go back to 2024 and then Poulpe loads :/

Fixed!

ortoLover (approved):
Everything looks good. From the code to testing functionality, everything is perfect and perfectly functional!!

PR #319, opened 2024-05-16
Create budget Item Screen

ortoLover:
We already have the edit budget item screen, and we want to use it to create a create budget item screen. The objective of this screen would be of creating a new budget item and push it on the database.

- [x] Create the UI
- [x] Create the appearance and disappearance of the screen
- [x] Create the functionality


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+import androidx.compose.ui.text.input.KeyboardType
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import androidx.compose.ui.window.Dialog
+import androidx.compose.ui.window.PopupProperties
+import com.github.se.assocify.model.entities.BudgetItem
+import com.github.se.assocify.model.entities.TVA
+import java.time.Year
+import java.util.UUID
+
```
sarahbadr17, line 40: java docs please

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+            supportingText = {})
+        Row(
+            modifier = Modifier.fillMaxWidth().padding(15.dp),
+            horizontalArrangement = Arrangement.SpaceBetween,
+        ) {
+          Button(
+              onClick = { budgetViewModel.cancelEditing() },
+              modifier = Modifier.padding(15.dp).testTag("editDismissButton"),
+          ) {
+            Text("Dismiss")
```
sarahbadr17, line 146: Either modify the name to cancel button or add an "cross icon button" `Icons.Default.Close` on the top right of the pop up

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+                      BudgetItem(
+                          budget.uid,
+                          nameItem = nameString,
+                          amount = amountString.toInt(),
+                          tva = TVA.floatToTVA(tvaString.toFloat()),
+                          description = descriptionString,
+                          subcategoryUID = budget.subcategoryUID,
+                          year = yearString.toInt()))
+                } else {
+                  Log.e("Error", budget.toString())
```
sarahbadr17, line 161: Debut message left?

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+                          amount = amountString.toInt(),
+                          tva = TVA.floatToTVA(tvaString.toFloat()),
+                          description = descriptionString,
+                          subcategoryUID = budget.subcategoryUID,
+                          year = yearString.toInt()))
+                }
+              },
+              modifier = Modifier.padding(15.dp).testTag("editConfirmButton"),
+          ) {
+            Text("Confirm")
```
sarahbadr17, line 175: For me, on a small phone, the buttons are not shown. If I make the Card scrollable: `Card(
        modifier = Modifier
            .padding(16.dp)
            .testTag("editDialogBox")
            .verticalScroll(rememberScrollState()),
        shape = RoundedCornerShape(16.dp),
    )`, they appear because I can scroll down to them. However, there's a difference of vertical spacing between these buttons and the rest of the outlined text field

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
```diff
@@ -0,0 +1,181 @@
...
+  BudgetPopUpScreen(budgetViewModel)
+}
+
+/**
+ * Displays the popup to edit a specific budget element
+ *
+ * @param budgetViewModel the viewModel of the budget details
+ */
+@Composable
+fun BudgetPopUpScreen(budgetViewModel: BudgetDetailedViewModel) {
```
sarahbadr17, line 57: Bug: I can create a budget Item without a name

sarahbadr17 (changes requested):
Some little fixes need otherwise good!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: Added it!!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: Modified to cancel, if I have time later will do a PR to transform to a cross button like in the balance screen

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: Removed it!!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: Changed ideas, I removed the cancel and replaced it with a delete only if there is edited. I also added the cross on the upper corner

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: Ok I have replaced the Column with a LazyColumn so the scrolling should be implemented now. For the spacing, I modified the buttons and now it should look good. I also modified a bit the rest of the UI for it to look better. If you have another comment to do don't hesitate!!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetPopUpScreen.kt
ortoLover: I have modified with all the modification. Now both for the amount and the name there are checks, and the year is done as a list. Hope that all your concerns were adressed!!

Mai-LinhC:
Testing on device : 
in budget :
- should put a placeholder for the description because if no description is put when creating an item the title/name looks like it's floating 
- need to use priceUtils because the amounts are int and not well handled yet
- the tva is calculated for the total only, not for the item : created an item with amount "100" (=1.00), put 8.1% tva, HT item and total are 1.00, TTC item is 1.00 and total is 1.08
- buggy dialog when too long description is being input (idk if we can do anything about that?)
- no restraints on the length of description so it's displayed fully, might want to restrict it
- every input is not single line (name, amount...) which is a bit annoying
in balance : 
- Status is named Tag instead :/
- uhhh i don't know what happened but i was editing an item, clicked on the X, tried to add something else by clicking the + and the app crashed

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetDetailedViewModel.kt
```diff
@@ -195,6 +223,53 @@ class BudgetDetailedViewModel(
...
+    _uiState.value = _uiState.value.copy(creating = false)
+  }
+
+  /**
+   * Checks if the format of title is good and changes the error depending on it
+   *
+   * @param title the string that is checked
+   */
+  fun setTitle(title: String) {
+    _uiState.value = _uiState.value.copy(titleError = title.isEmpty() || title.length > 20)
```
Polymeth, line 255: nit: magic value here

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetDetailedViewModel.kt
```diff
@@ -195,6 +223,53 @@ class BudgetDetailedViewModel(
...
+   */
+  fun setAmount(amount: String) {
+    _uiState.value =
+        _uiState.value.copy(
+            amountError =
+                amount.isEmpty() || amount.toDoubleOrNull() == null || amount.toDouble() < 0.0)
+  }
+
+  fun setDescription(description: String) {
+    _uiState.value = _uiState.value.copy(descriptionError = description.length > 50)
```
Polymeth, line 271: nit: magic value here too

Polymeth (commented):
Tried to torture the screen as much as possible and after all of those modificatons, it looks pretty good ! 

Only two "problems" remains for me : 
- when you put a long description you can still add the item. It says "description too long" but it's still added and then displayed with the too long description.
- the "+" button is displayed over the total when the list of items is too long, at some point. we could add padding like in other screens

ortoLover:
Responded to the problems you gave. Now it should work.

Mai-LinhC:
Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)

Mai-LinhC:
created an item for 2021 while having the filter 2024, it showed the new item (when it shouldn't). going to 2021 filter it does show the item and going back to 2024 it doesn't show. but should probably fix that it shouldn't necessarily be added to the list of shown items, i'd say :)

Mai-LinhC:
can't delete an item

Mai-LinhC:
Can't dismiss the dialog to create item ?

Mai-LinhC:
If there was an error previously (description too long...) and you quit the dialog, the error is not cleared (you open the dialog and the error is still here)

ortoLover:
> Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)

I misinterpreted a comment that you made then I thougth that was a placeholder. I will then remove the initial description.

ortoLover:
> can't delete nor edit an item ?

This is a problem with the database. I can fix it if you want but maybe it's better to do it on another PR.

Mai-LinhC:
> > Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)
> 
> I misinterpreted a comment that you made then I thougth that was a placeholder. I will then remove the initial description.

My bad I thought you understood, I meant to say, like in receipt, if there's no description, we need to display a placeholder so that it looks good because otherwise the name just looks not aligned

Mai-LinhC:
> > can't delete nor edit an item ?
> 
> This is a problem with the database. I can fix it if you want but maybe it's better to do it on another PR.

alrighty no problem, i'm adding an issue in the backlog then

ortoLover:
> Can't dismiss the dialog to create item ?

Can you give me in what conditions it does not work? There is a cross in the rigthmost screen to dismiss
Screenshot from 2024-05-19 00-02-57

Mai-LinhC:
> > Can't dismiss the dialog to create item ?
> 
> Can you give me in what conditions it does not work? There is a cross in the rigthmost screen to dismiss Screenshot from 2024-05-19 00-02-57

I mean that you can't dismiss by clicking outside of the dialog, which is usually possible with dialogs :) but yeah it's fine otherwise since there's the cross, just weird since all the other dialogs are dismissable this way

ortoLover:
> If there was an error previously (description too long...) and you quit the dialog, the error is not cleared (you open the dialog and the error is still here)

Fixed

ortoLover:
> > > Can't dismiss the dialog to create item ?
> > 
> > 
> > Can you give me in what conditions it does not work? There is a cross in the rigthmost screen to dismiss Screenshot from 2024-05-19 00-02-57
> 
> I mean that you can't dismiss by clicking outside of the dialog, which is usually possible with dialogs :) but yeah it's fine otherwise since there's the cross, just weird since all the other dialogs are dismissable this way

fixed

ortoLover:
> created an item for 2021 while having the filter 2024, it showed the new item (when it shouldn't). going to 2021 filter it does show the item and going back to 2024 it doesn't show. but should probably fix that it shouldn't necessarily be added to the list of shown items, i'd say :)

Yes there is a problem with a list of shown years that is different from the list of year that is current, just didn't think I had to correct it in this PR. Still I can correct it if you want.

Mai-LinhC:
> > created an item for 2021 while having the filter 2024, it showed the new item (when it shouldn't). going to 2021 filter it does show the item and going back to 2024 it doesn't show. but should probably fix that it shouldn't necessarily be added to the list of shown items, i'd say :)
> 
> Yes there is a problem with a list of shown years that is different from the list of year that is current, just didn't think I had to correct it in this PR. Still I can correct it if you want.

not sure what you mean but imma check the code to see if i see the problem i think is occuring, i'll ping you in a sec

ortoLover:
> Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)

In my case the empty description works. However fixed the description text.

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetDetailedViewModel.kt
```diff
@@ -195,6 +225,57 @@ class BudgetDetailedViewModel(
...
+    if (_uiState.value.titleError ||
+        _uiState.value.amountError ||
+        _uiState.value.descriptionError) {
+      return
+    }
+    budgetApi.addBudgetItem(
+        CurrentUser.associationUid!!,
+        budgetItem,
+        {
+          _uiState.value = _uiState.value.copy(budgetList = _uiState.value.budgetList + budgetItem)
```
Mai-LinhC, line 260: yup found the "problem" : you added the new budget item by default, even if it shouldn't be shown because the filters are not the right ones

Mai-LinhC:
> > Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)
> 
> In my case the empty description works. However fixed the description text.

Yup it works my bad, it was because i was editing it and that doesn't work !

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/budget/BudgetDetailedViewModel.kt
ortoLover: Ok thanks will fix it

ortoLover:
> > > Description is by default "Description" and we have to delete it to be able to write, should probably remove that + Can't create empty description (just doesn't work, i think it should but if you absolutely want a description there should be an error displayed to tell that it needs one)
> > 
> > 
> > In my case the empty description works. However fixed the description text.
> 
> Yup it works my bad, it was because i was editing it and that doesn't work !

Ok I checked and it's due to the database error. I will change it because it's important and everyone always asks about it.

sarahbadr17 (approved):
Looks good to me! You can merge

PS: the app crashes when I click on the add button on the balance detailed screen. I guess it's fixed in the other PR, if it's not please take that into consideration.

PR #320, opened 2024-05-16
BugFix : cache problem when joining asso

Mai-LinhC:
Currently there's a problem when joining an association because we're trying to get the information from the cache except it's not yet in the cache since we're just joining...

### DIFF
--- a/app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
+++ b/app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
@@ -173,10 +173,17 @@ class CreateAssociationViewmodel(
     assoAPI.addAssociation(
         association,
         onSuccess = {
-          assoAPI.initAssociation(roles.values, _uiState.value.members, {}, {})
           CurrentUser.associationUid = association.uid
-          userAPI.updateCurrentUserAssociationCache({}, {})
-          navActions.goFromCreateAsso()
+          assoAPI.initAssociation(
+              roles.values,
+              _uiState.value.members,
+              {
+                userAPI.updateCurrentUserAssociationCache(
+                    {   Log.e("lalala", "update current : ${it}")
+                        navActions.goFromCreateAsso() },
+                    { Log.e("error", "update current : ${it.message}") })
+              },
+              { Log.e("error", "init asso : ${it.message}") })
         },
         onFailure = { exception ->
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
```diff
@@ -178,9 +178,14 @@ class CreateAssociationViewmodel(
...
+              roles.values,
+              _uiState.value.members,
+              {
+                CurrentUser.associationUid = association.uid
+                userAPI.updateCurrentUserAssociationCache({ navActions.goFromCreateAsso() }, {})
+              },
+              {})
         },
         onFailure = { exception ->
           Log.e("CreateAssoViewModel", "Failed to add asso: ${exception.message}")
```
SidonieBouthors, line 191: Maybe adding at least some error logging for each onError would be better (or even error message / condition for what to do if relevant)

app/src/main/java/com/github/se/assocify/ui/screens/selectAssociation/SelectAssociationViewModel.kt
```diff
@@ -49,26 +50,43 @@ class SelectAssociationViewModel(
...
+                              }
+                            },
+                            {})
+                      },
+                      {})
                 }
               },
               {})
         },
         {}) // TODO: handle joining assoc (non immediate approval)
```
SidonieBouthors, line 89: Make sure to remove debug error logs when the operation is a success :) And same as previously, adding error logging / messages for actual errors

SidonieBouthors (commented):
Needs some more attention to dealing with errors, but the fix works great on my end, both for joining and creating a new assoc :)) Great work !

app/src/main/java/com/github/se/assocify/ui/screens/createAssociation/CreateAssociationViewmodel.kt
SekoiaTree: There wasn't any previously, we're gonna add it in the next sprint

app/src/main/java/com/github/se/assocify/ui/screens/selectAssociation/SelectAssociationViewModel.kt
SekoiaTree: Fixed!

SidonieBouthors (approved):
Looks good to me, works as expected, great bugfix :)

PR #329, opened 2024-05-17
Treasury Loading

SidonieBouthors:
- [x] Implement loading + error message for main accounting
- [x] Implement loading + error message for sub accounting (details)
- [x] Implement error snackbars


app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingViewModel.kt
```diff
@@ -85,11 +110,13 @@ class AccountingViewModel(
     // if the category is global, display all subcategories
     if (categoryName == "Global") {
       _uiState.value = _uiState.value.copy(globalSelected = true)
+      loadAccounting()
```
sarahbadr17, line 113: nit: since loadAccounting is called in both cases you could just add it once at the end of the cases

sarahbadr17, line 113: nit: since loadAccounting is called in both cases you could just add it once at the end of the cases

sarahbadr17 (approved):
LGTM ! You did a wonderful job. Thank you !

PR #332, opened 2024-05-17
Make text entries single line

### LINKED ISSUE #331: Text inputs are multiline

SekoiaTree:
Closes #331

Previously, many text entries were multi-line. This made it possible to have multi-line names, which look strange and might break the UI. This fixes that. It is local-only, but because it only causes visual glitches, it's okay that it's not in the database.


Mai-LinhC (approved):
LGTM

PR #335, opened 2024-05-17
Search categories

sarahbadr17:
This PR implements the search of subcategories. Plus, I modified the accountingViewmodel so that we fetch only once the list of subcategory, and use the resulting local list to filter.

- [x] Add function in viewmodel
- [x] Modify function to get subcategory list in AccountingViewModel
- [x] Modify UI screen
- [x] Tests

### DIFF
--- a/app/src/main/java/com/github/se/assocify/ui/composables/MainTopBar.kt
+++ b/app/src/main/java/com/github/se/assocify/ui/composables/MainTopBar.kt
@@ -74,6 +74,7 @@ fun MainTopBar(
                   searchBarVisible = true
                 }) {
                   Icon(imageVector = Icons.Filled.Search, contentDescription = "Search")
+                  Icon(imageVector = Icons.Filled.Search, contentDescription = "Search")
                 }
           }
         },
@@ -87,7 +88,14 @@ fun MainTopBar(
         onSearch = onSearch,
         active = false,
         onActiveChange = {},
-        placeholder = { Text("Search receipts") },
+        placeholder = {
+            when(page){
+                0 -> Text("Search receipts")
+                1 -> Text("Search accounting categories")
+                2 -> Text("Search accounting categories")
+                else -> Text("Search")
+            }
+        },
         leadingIcon = {
           IconButton(
               modifier = Modifier.testTag("searchBackButton"),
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/github/se/assocify/ui/composables/MainTopBar.kt
```diff
@@ -74,6 +74,7 @@ fun MainTopBar(
                   searchBarVisible = true
                 }) {
                   Icon(imageVector = Icons.Filled.Search, contentDescription = "Search")
+                  Icon(imageVector = Icons.Filled.Search, contentDescription = "Search")
```
SidonieBouthors, line 77: Accidental duplicated line here I think :)

app/src/main/java/com/github/se/assocify/ui/composables/MainTopBar.kt
```diff
@@ -87,7 +88,14 @@ fun MainTopBar(
...
         onActiveChange = {},
-        placeholder = { Text("Search receipts") },
+        placeholder = {
+          when (page) {
+            0 -> Text("Search receipts")
+            1 -> Text("Search accounting categories")
+            2 -> Text("Search accounting categories")
+            else -> Text("Search")
+          }
+        },
```
SidonieBouthors, line 98: The original hardcoded text must have been a mistake on my part, we definitely should pass this as argument :)

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingViewModel.kt
```diff
@@ -26,41 +26,56 @@ class AccountingViewModel(
...
-                _uiState.value.copy(
-                    subCategoryList =
-                        subCategoryList.filter {
-                          it.categoryUID == _uiState.value.selectedCatUid &&
-                              it.year == _uiState.value.yearFilter
-                        })
-          }
+          _uiState.value = _uiState.value.copy(allSubCategoryList = subCategoryList)
         },
         { Log.d("BudgetViewModel", "Error getting subcategories") })
```
SidonieBouthors, line 52: Probably should remove the debug log, but this will be fixed in my PR anyways

SidonieBouthors (commented):
Good PR :) The search works great ! Only a couple minor things to change

SidonieBouthors (approved):
Great fix :) Thanks for the PR

Issue #334, opened 2024-05-17
Adapt profile to what will be functional


Mai-LinhC:
Change the profile settings : adapt it to what we're actually going to implement.
What will be here : 
- Preferences 
- Members
- Treasury
- Events

Therefore we're removing from view : 
- Privacy/Security (not in V1)
- Notifications (won't be done in V1)
- Roles (they're predefined in V1)

PR #337, opened 2024-05-17
Adapt profile to what will be functional

Mai-LinhC:
For the issue #334 : 

> Change the profile settings : adapt it to what we're actually going to implement.
> What will be here :
> 
> Preferences
> Members
> Treasury
> Events
>
> Therefore we're removing from view :
> 
> Privacy/Security (not in V1)
> Notifications (won't be done in V1)
> Roles (they're predefined in V1)


SidonieBouthors (commented):
We may want to consider completely removing things that won't be implemented for V1 rather than commenting, in order to avoid some code fossils remaining in our app :) To discuss with the team

SidonieBouthors (approved):
Looks good to me! Good job :)

PR #341, opened 2024-05-18
Add a license to the project 

zizouet:
The GNU General Public License version 3 (GPLv3) is a free software license that ensures users have the freedom to run, study, share, and modify the software. Here's a simple breakdown of its key points:

Freedom to Use: You can use the software for any purpose without restrictions.

Access to Source Code: You have the right to access the software's source code, which allows you to study how it works and modify it.

Freedom to Share: You can redistribute copies of the software to others, either in its original form or with your modifications.

Share Alike: If you distribute modified versions of the software, you must also distribute the source code and keep it under the same GPLv3 license. This ensures that future users also have the same freedoms.

Protection Against Tivoization: GPLv3 includes provisions against "tivoization," which refers to hardware restrictions that prevent users from running modified versions of the software on certain devices.

Patent Rights: The license includes protections against patent lawsuits, ensuring that users are not threatened by patents when using, modifying, or distributing the software.

Compatibility with Other Licenses: GPLv3 is designed to be compatible with other free software licenses, making it easier to integrate code from different sources.

In summary, GPLv3 ensures that software remains free and open, protecting users' rights to use, study, share, and modify the software, while preventing restrictions that could undermine these freedoms.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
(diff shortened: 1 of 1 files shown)


sarahbadr17 (approved):
Well thought!

PR #344, opened 2024-05-18
Fix Task API cache

zizouet:
Task api wasn't updating the cache when deleting a task. Now it is updated :)


ortoLover (approved):
Everything looks good to me!

PR #346, opened 2024-05-18
Fix the TVA issues on single item

ortoLover:
There is a little problem: the TVA on the single values is not activated when the filter of TVA is activated. The objective of this PR is to fix this single issue so it can be repaired and the TVA can be consistent everywhere.


Mai-LinhC (approved):
LGTM !

PR #358, opened 2024-05-19
Repair some errors that remain

ortoLover:
There were some problems remaining in some screens. This PR is to address those problems and correct functionality issues.


sarahbadr17 (approved):
Don't see any bugs. Thanks

PR #364, opened 2024-05-19
Repair APK CI

SekoiaTree:
We forgot to fix the CI after M2. This PR brings that fix to the main branch. It only fixes the APK generation CI, because the testing CI doesn't need to have access to the actual database, and so it is ensured.


Mai-LinhC (approved):
LGTM, can log in

PR #367, opened 2024-05-22
Add receipt owners to receipt entity

### LINKED ISSUE #306: Fix : No user in Receipt entity

SekoiaTree:
Closes #306.

This PR adds a `userId` field to `Receipt`, which already existed in the database but was simply not linked. This PR also makes editing receipts not transfer their owner.


app/src/main/java/com/github/se/assocify/ui/screens/treasury/receiptstab/receipt/ReceiptViewModel.kt
```diff
@@ -43,6 +46,8 @@ class ReceiptViewModel {
     this.navActions = navActions
     this.receiptApi = receiptApi
     this.receiptUid = receiptUid
+    // Temporary. Gets overridden once the receipt is fetched
+    this.receiptCreatorUid = CurrentUser.userUid!!
```
Mai-LinhC, line 50: Is it necessary to set it since it's going to be overwritten ?

Mai-LinhC (commented):
Just a comment otw LGTM

Mai-LinhC (approved):
The comment makes sense and there should be no problem with the temporary set, all good

PR #371, opened 2024-05-23
Fetch members in association API

### LINKED ISSUE #370: Add a way to fetch members of an association
This allows listing all members of an association. Preferably this method would also fetch the roles of each user, and the users, at the same time.

SekoiaTree:
Closes #370. Adds a method in the assocation API to fetch members of an association.

- If the association is not present in the cache, the cache is updated (just in case). If it's still not present, we fail. The association cache not being up-to-date enough for this is rare, but one extra API call doesn't cost much so...
- There is a basic cache that stores 1 member list (generally, we don't request different member lists very often)
- The cache is invalidated when the main association cache is refreshed, mostly because adding another method to call makes `AssociationAPI` more complex and annoying to use.


app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -260,6 +261,56 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+            db.from("member_of")
+                .select(Columns.list("users(*)", "role!inner(*)")) {
+                  filter { eq("role.association_id", associationId) }
+                }
+                .decodeList<MemberOf>()
+                .map { AssociationMember(it.user, association, it.role) }
+        memberCache = associationId to members
+        onSuccess(members)
+      }
+    } else {
```
Mai-LinhC, line 303: I don't get why the cache is refreshed only in this function and not in others. In other functions, if the association is not in the cache you just fail. I think it's good that it's refreshed but I don't understand why only here haha

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -260,6 +261,56 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
...
+    val association = associationCache[associationId]
+    if (association != null) {
+      tryAsync(onFailure, tag = "getMembers") {
+        val members =
+            db.from("member_of")
+                .select(Columns.list("users(*)", "role!inner(*)")) {
+                  filter { eq("role.association_id", associationId) }
+                }
+                .decodeList<MemberOf>()
+                .map { AssociationMember(it.user, association, it.role) }
```
Mai-LinhC, line 299: is it not possible to directly have it as an AssociationMember ?

app/src/test/java/com/github/se/assocify/model/database/AssociationAPITest.kt
```diff
@@ -400,4 +399,60 @@ class AssociationAPITest {
...
+    assoAPI.getMembers(APITestUtils.ASSOCIATION.uid, successMock, failureMock)
+    verify(timeout = 1000) { successMock(listOf(APITestUtils.ASSOCIATION_MEMBER)) }
+    verify(exactly = 0) { failureMock(any()) }
+    clearMocks(successMock, failureMock)
+
+    error = true
+    assoAPI.getMembers(APITestUtils.ASSOCIATION.uid, successMock, failureMock)
+    verify(timeout = 1000) { successMock(listOf(APITestUtils.ASSOCIATION_MEMBER)) }
+    verify(exactly = 0) { failureMock(any()) }
+    clearMocks(successMock, failureMock)
```
Mai-LinhC, line 449: Why does it succeed here when you put error to true ?

SekoiaTree: This part tests cache.

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
```diff
@@ -260,6 +261,56 @@ class AssociationAPI(private val db: SupabaseClient) : SupabaseApi() {
     }
   }
 
+  @Serializable
+  private data class MemberOf(
+      @SerialName("users") val user: User,
+      @SerialName("role") val role: PermissionRole,
+  )
```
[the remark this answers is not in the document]
SekoiaTree: `AssociationMember` includes an association, so I'd need to join with that table. This is faster.

app/src/main/java/com/github/se/assocify/model/database/AssociationAPI.kt
Mai-LinhC: I mean that the main cache is updated only in this one. If you check the other functions in this API, when the association is not found in the (main) cache, we just call onFailure instead of doing, like in this function, a refresh of the cache and then a failure if the refresh was unsuccessful

SekoiaTree: Good point. However, I can't imagine a case where you'd have access to an association ID without it being in the cache. This way, failure is "instant", instead of taking a while to figure out. I think I prefer simply not refreshing.

PR #374, opened 2024-05-24
Extract image loading into class

SekoiaTree:
Part of #338 and #262. This PR moves image loading from `ReceiptAPI` into a separate class. It also adds image loading APIs to both relevant APIs, but does not connect it with the viewmodels. That will be done in a separate PR.


SidonieBouthors (approved):
Looks good, and had some nice features to avoid issues (tmp images when fetching, use of nonces...). Too bad it is so difficult to test some parts of this, we'll have to test manually with the other PRs. Good job !

PR #379, opened 2024-05-25
Profile pictures for users

### LINKED ISSUE #262: Profile pictures

SekoiaTree:
Closes #262. This PR adds the ability to add and load profile pictures. If the image cannot be fetched, the default icon is displayed and an error is shown, but the page still loads, because it is not a core functionality.

Still needs to be done: refresh the profile picture on reload.

### DIFF
--- a/app/src/main/java/com/github/se/assocify/ui/screens/profile/ProfileViewModel.kt
+++ b/app/src/main/java/com/github/se/assocify/ui/screens/profile/ProfileViewModel.kt
@@ -85,6 +85,15 @@ class ProfileViewModel(
           endLoading()
         },
         { endLoading("Error loading profile") })
+    userAPI.getProfilePicture(
+        "${CurrentUser.userUid!!}.jpg",
+        { uri -> _uiState.value = _uiState.value.copy(profileImageURI = uri) },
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/se/assocify/ui/screens/profile/ProfileViewModel.kt
```diff
@@ -87,6 +91,15 @@ class ProfileViewModel(
           endLoading()
         },
         { endLoading("Error loading profile") })
+    userAPI.getProfilePicture(
+        "${CurrentUser.userUid!!}.jpg",
+        { uri -> _uiState.value = _uiState.value.copy(profileImageURI = uri) },
+        {
+          CoroutineScope(Dispatchers.Main).launch {
+            _uiState.value.snackbarHostState.showSnackbar(
+                message = "Error loading profile picture", duration = SnackbarDuration.Short)
```
ortoLover, line 100: Maybe add specific error for when the image cannot be found?

ortoLover (approved):
The PR looks good. I tested it on the emulator and it looks like it works well.  Good job!

PR #380, opened 2024-05-26
Decrease AccountingDetailedScreen complexity

ortoLover:
We noticed that this screen had a complexity that was too big. The objective of this PR is to reduce it so that the code remains readable in the years to come.

### DIFF
--- a/app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
+++ b/app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
@@ -1,6 +1,5 @@
 package com.github.se.assocify.model.entities
 
-import androidx.compose.material3.Text
 import java.time.LocalDate
 
 /**
@@ -28,11 +27,11 @@ data class BalanceItem(
     val date: LocalDate,
     val assignee: String,
     val status: Status,
-){
-    fun getAmount(tvaActive: Boolean): Int{
-        return  if (tvaActive) this.amount + (this.amount * this.tva.rate / 100f).toInt() else this.amount
-    }
-
+) {
+  fun getAmount(tvaActive: Boolean): Int {
+    return if (tvaActive) this.amount + (this.amount * this.tva.rate / 100f).toInt()
+    else this.amount
+  }
 }
 
 /**
--- a/app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
+++ b/app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
@@ -1,5 +1,6 @@
 package com.github.se.assocify.model.entities
 
+import androidx.compose.material3.Text
 import java.time.LocalDate
 
 /**
@@ -27,7 +28,12 @@ data class BalanceItem(
     val date: LocalDate,
     val assignee: String,
     val status: Status,
-)
+){
+    fun getAmount(tvaActive: Boolean): Int{
+        return  if (tvaActive) this.amount + (this.amount * this.tva.rate / 100f).toInt() else this.amount
+    }
+
+}
 
 /**
  * Represents the TVA of a budget or balance item, these types are also represented in the database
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
```diff
@@ -27,7 +27,12 @@ data class BalanceItem(
     val date: LocalDate,
     val assignee: String,
     val status: Status,
-)
+) {
+  fun getAmount(tvaActive: Boolean): Int {
```
sarahbadr17, line 31: Well thought!

app/src/main/java/com/github/se/assocify/model/entities/BudgetItem.kt
```diff
@@ -18,4 +18,13 @@ data class BudgetItem(
...
     val subcategoryUID: String,
     val year: Int
-)
+) {
+  fun getAmount(tvaActive: Boolean): Int {
+    return if (tvaActive) this.amount + (this.amount * this.tva.rate / 100f).toInt()
+    else this.amount
+  }
+
+  fun getFormattedDescription(): String {
```
sarahbadr17, line 27: why don't you have the same function for balanceItem too?

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -110,15 +90,10 @@ fun AccountingDetailedScreen(
...
-                    when (page) {
-                      AccountingPage.BALANCE ->
-                          if (balanceState.subCategory != null) {
-                            balanceDetailedViewModel.startSubCategoryEditingInBalance()
-                          }
-                      AccountingPage.BUDGET ->
-                          if (budgetState.subCategory != null) {
-                            budgetDetailedViewModel.startSubCategoryEditingInBudget()
-                          }
+                    if (page == AccountingPage.BALANCE && balanceState.subCategory != null) {
```
sarahbadr17, line 93: you can replace balanceState.subCategory by subCategory since it's passed as a parameter of the function

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
```diff
@@ -195,7 +160,6 @@ fun AccountingDetailedScreen(
               }
             }
           }
-
           // Display the items
           when (page) {
```
sarahbadr17, line 164: Is there a way to avoid this when(page), they both do the same thing, the only parameter that changes is the list and filterACtive. 

Maybe you can create an external function?

sarahbadr17 (commented):
Just 2-3 comments! Very good! Hope it suffices. I tested it and the functionality didn't change. Well job :)

app/src/main/java/com/github/se/assocify/model/entities/BalanceItem.kt
ortoLover: Thank you!!

app/src/main/java/com/github/se/assocify/ui/screens/treasury/accounting/AccountingDetailedScreen.kt
ortoLover: I will think of a way to repair this

ortoLover: I tried building an auxiliary function, but it had too many parameters. Thus I prefer keeping it this way. We can modify it further if we see that the complexity still is higher than 50.

sarahbadr17 (approved):
Looks good to me! Hope it suffices to decrease the complexity

PR #381, opened 2024-05-26
Profile pictures for associations

SekoiaTree:
Part of #338. Adds association logo creation, but does not connect it to the top bar logo because it is separate.


ortoLover (commented):
There is a problem: the profile picture does not save and does not work on the emulator

ortoLover (approved):
THe error was corrected in the database, everything is good now good job!!

PR #386, opened 2024-05-29
Remove the Home and Chat screen

ortoLover:
Now that we have the final submission we have to start cutting the functionalities that won't make it until the end. This PR has the objective to cut everything that was related to the Home and the Chat screen now that we decided that the final version of the app won't use it.


Polymeth (approved):
Perfect LGTM :)

PR #388, opened 2024-05-29
Link event to asso

zizouet:
Changing associations now change the events you're seeing in real-time, not only at the beginning of the app.


SidonieBouthors (approved):
Looks good to me, good fix :)

PR #389, opened 2024-06-01
Permissions of receipt

sarahbadr17:
This PR displays the receipt list depending on the role the current user has. If the user is a president or a treasurer, they can see every receipt (their receipts and all receipts), if the user has any other role, they can only see their own receipts.


app/src/main/java/com/github/se/assocify/ui/screens/treasury/receiptstab/ReceiptListViewModel.kt
```diff
@@ -109,5 +127,7 @@ data class ReceiptUIState(
     val error: String? = null,
     val userReceipts: List<Receipt> = listOf(),
     val allReceipts: List<Receipt> = listOf(),
-    val searchQuery: String = ""
+    val searchQuery: String = "",
+    val userCurrentRole: PermissionRole =
+        PermissionRole(CurrentUser.userUid!!, CurrentUser.associationUid!!, RoleType.MEMBER)
```
zizouet, line 132: I think member role type as default is dangerous, first there exist "staff" roletype that is less privileged than a member. Also there should be no default roletype except a Roletype with no privilege which actually don't exist in our implementation. Maybe either create it or let no default userCurrentRole :)

zizouet (approved):
LGTM to me, works well on my phone, i can't see any of the receipts of other even using search bar :))

PR #392, opened 2024-06-01
Add a ViewModel to TreasuryTag screen

ortoLover:
The parameter tab is full of sub-tabs, and one of the objectives of this pr is to add the ability to add and edit a tag that will be used in treasury. Thus the PR will focus on the creation of a viewModel to edit and add a new treasury tag


app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+    navActions: NavigationActions
+) : ViewModel() {
+  private val _uiState = MutableStateFlow(ProfileTreasuryTagsUIState())
+  val uiState: StateFlow<ProfileTreasuryTagsUIState> = _uiState
+
+  init {
+    accountingCategoryAPI.getCategories(
+        CurrentUser.associationUid!!,
+        { categoryList -> _uiState.value = _uiState.value.copy(treasuryTags = categoryList) },
+        { Log.e("treasurytags", "Error loading tags") })
```
SidonieBouthors, line 23: We might want to consider adding the usual loading + error message system maybe?

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsViewModel.kt
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  fun deleteTag(tag: AccountingCategory) {
+    accountingCategoryAPI.deleteCategory(
+        tag,
+        {
+          _uiState.value =
+              _uiState.value.copy(treasuryTags = _uiState.value.treasuryTags.filter { tag != it })
+        },
+        {})
```
SidonieBouthors, line 43: It would be nice to have snackbars for these when adding or deleting a tag fails

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsScreen.kt
```diff
@@ -0,0 +1,164 @@
...
+            item {
+              OutlinedTextField(
+                  singleLine = true,
+                  isError = nameString.isEmpty(),
+                  modifier = Modifier.padding(8.dp).testTag("nameField"),
+                  value = nameString,
+                  onValueChange = { nameString = it },
+                  label = { Text("Name") },
+                  supportingText = {
+                    Text(if (nameString.isEmpty()) "The string should not be empty!" else "")
```
SidonieBouthors, line 141: Consider making the error come from the state, that way you can set it only if the text field becomes empty by user action (or save is pressed), rather than as soon as the popup shows up

SidonieBouthors (commented):
Overall works great on my end and everything seems to go as expected. Just a few notes on dealing with errors and loading in a more user friendly way :)

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsViewModel.kt
ortoLover: Ok will do it now

SidonieBouthors: Make sure the top bar with the back button is still accessible :) you can use error page instead of error message if needed. Also, the error message currently says "Error loading receipts"

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsScreen.kt
SidonieBouthors: Nice :) It works now

SidonieBouthors (approved):
Thanks for the loading and error handling :) looks good to me !

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsViewModel.kt
SidonieBouthors: Thanks!

app/src/main/java/com/github/se/assocify/ui/screens/profile/treasuryTags/ProfileTreasuryTagsViewModel.kt
SidonieBouthors: Works great now :)

PR #395, opened 2024-06-01
Assoc Settings Perms

SidonieBouthors:
Makes the association settings only visible to committees, president or treasurer of an assoc.


Polymeth (approved):
Perfect PR well done :)

PR #402, opened 2024-06-02
Bug fix: amount in budget can be negative

sarahbadr17:
This PR enables the fact that amount in budgets can be negative. Of course, you can budget an expense... There was just a check in the viiewmodel that I removed.


SekoiaTree (approved):
LGTM

PR #403, opened 2024-06-02
Permission for budget and balance

sarahbadr17:
Only the treasurer or the president can see the accounting books. Any other member only sees the receipts in the tab treasury.


Polymeth (approved):
Everything tested, everything works. Thanks for always being a great programmer :)

PR #407, opened 2024-06-02
Methods for managing Association Members

zizouet:
Added the possibility to remove the user from an association.
Added the possibility to change the role of a user in an association.

Enjoy :)))


zizouet:
Still not good :((

PR #410, opened 2024-06-02
Added GPS location on the map

Polymeth:
The PR adds a little character on the GPS location of the user. 
NB : only testable on phone. Can be tested on emulator by using location settings of it but the emulator isn't really good at simulating GPS.

This PR addresses #326.


zizouet (approved):
Wonderful it works perfectly on my phone, moving makes it move and it shows my real position !
Sadly OSMdroid is not as good as maps :(

PR #411, opened 2024-06-02
Fix last Sonarcloud issues

ortoLover:
We want to fix some issues that sonarcloud gives us to improve the quality of our code.


ortoLover:
I just noticed it was not important, aborting the PR

PR #414, opened 2024-06-02
Fix map right/left slide

Polymeth:
Because the component for tab change, a bug re-appeared. When on the map tab, sliding left/right was sliding tab instead of sliding on the map which is very annoying for user experience.

This PR addresses #413


ortoLover (approved):
🚀

PR #415, opened 2024-06-02
Remove all todos

ortoLover:
We had the bad practice of leaving todos in the middle of the code. Now it's too late to keep them so we decided to remove them.


SekoiaTree (approved):
LGTM

PR #417, opened 2024-06-02
Current association Image

zizouet:
You can now see the current association you're in in the upper left of the screen !
>:)


app/src/main/java/com/github/se/assocify/ui/composables/MainTopBar.kt
```diff
@@ -59,11 +73,29 @@ fun MainTopBar(
...
         title = { Text(text = title) },
         navigationIcon = {
-          IconButton(modifier = Modifier.testTag("accountIconButton"), onClick = {}) {
-            Icon(
-                imageVector = Icons.Filled.AccountCircle,
-                contentDescription = "Association Account")
-          }
+          IconButton(
+              modifier = Modifier.testTag("accountIconButton"),
+              onClick = { /*TODO On assoc account click */}) {
```
ortoLover: There is a todo here

ortoLover (approved):
Tested it on the emulator it work, good job!!

